PR #138, opened 2022-09-23
Feature/bngp 817 npm publish

teddmason:


teddmason:
@reneli5 work includes a script that should version and tag merges into master (develop in the future), and then publish the modules to npm.

Also includes work to remove the workarounds for the symlink-ed local modules that can now be pulled from npm.

1 caveat is that until the script is in master it is difficult to debug/test, so may need some further dummy pull requests to get the script working right.

PR #119, opened 2022-09-07
Feature/bng 768 check landboundary

reneli5:
Updated preview land boundary to new design

PR #174, opened 2022-10-07
BNGP - 184 and 859 - landowner section pages

teddmason:


packages/application-to-register-webapp/webpack.config.js
```diff
@@ -30,13 +31,5 @@ export default {
...
     ]
-  },
-  plugins: [
-    new HtmlPlugin({
-      inject: false,
-      filename: '../views/land/confirm-land-boundary.html',
-      template: 'src/views/webpack/land/confirm-land-boundary.html',
-      chunks: ['confirmLandBoundary']
-    })
-  ]
```
pwadmore-ea, line 41: While I've been away has there been a decision to move away from the approach to handling page specific JavaScript through webpack as used by the Future Farming and Countryside programme?

It looks like script tags are now being placed in HTML files directly (which looks simpler). I've made a note that the page specific JavaScript associated with confirmation of a geospatial land boundary upload will need revisiting when this functionality is brought back into scope as it will be broken right now. It looks like the page specific HTML (and associated page specific JavaScript) was simply removed in  #119).

teddmason: Ye my choice, I've moved to how we did it on flood. The webpack method for the geospatial upload was an extra step that I see as being unnecessary. K.I.S.(S)

teddmason: And to note we still have page specific scripts processed through webpack

pwadmore-ea (dismissed):

pwadmore-ea (approved):

PR #182, opened 2022-10-12
static pages for service guidance kick out

teddmason:


packages/application-to-register-webapp/src/routes/__tests__/land/need-boundary-file.spec.js
```diff
@@ -0,0 +1,11 @@
...
+import constants from '../../../utils/constants.js'
+const url = constants.routes.NEED_BOUNDARY_FILE
+
+describe(url, () => {
+  describe('GET', () => {
+    it(`should render the ${url.substring(1)} view`, async () => {
+      await submitGetRequest({ url })
+    })
+  })
+})
```
teddmason, line 11: test

packages/application-to-register-webapp/src/routes/__tests__/land/need-legal-agreement.spec.js
```diff
@@ -0,0 +1,11 @@
+import { submitGetRequest } from '../helpers/server.js'
+import constants from '../../../utils/constants.js'
+const url = constants.routes.NEED_LEGAL_AGREEMENT
+
+describe(url, () => {
+  describe('GET', () => {
+    it(`should render the ${url.substring(1)} view`, async () => {
+      await submitGetRequest({ url })
```
teddmason, line 8: test2

pwadmore-ea (approved):
@teddmason - I'm happy to approve this. However, I think there is a a small amount of technical debt building concerning the maximum size of uploads. 50MB seems to be hardcoded into most (if not all )views and associated view preparation code.  If the maximum upload size changes for one or more uploads, all static references will need updating accordingly.

Additionally (from earlier pull requests), upload routes appear to use constants associated with the maximum file size of the original geospatial land boundary upload rather than having their own analogous constants.

I suggest adding low  priority technical debt tasks to the backlog accordingly.

pwadmore-ea:
> > @teddmason - I'm happy to approve this. However, I think there is a a small amount of technical debt building concerning the maximum size of uploads. 50MB seems to be hardcoded into most (if not all )views and associated view preparation code. If the maximum upload size changes for one or more uploads, all static references will need updating accordingly.
> > Additionally (from earlier pull requests), upload routes appear to use constants associated with the maximum file size of the original geospatial land boundary upload rather than having their own analogous constants.
> > I suggest adding low priority technical debt tasks to the backlog accordingly.
> 
> At the moment we haven't been given any value other than 50mb. So i don't see the need to add anything to the backlog, I think when the time comes that they change this value for any of the uploads then at that point we would need to update the code to allow, but until that time no need to waste time on it

Fair enough @teddmason - I'm happy for this pull request to be merged accordingly.

PR #185, opened 2022-10-21
Feature/bngp 738 bngp 739 upload and check metric file

AjinkyaJawalkar:


packages/application-to-register-webapp/src/routes/developer/check-metric-file.js
```diff
@@ -0,0 +1,60 @@
...
+            text: 'Select yes if this is the correct file',
+            href
+          }
+        ]
+      })
+    }
+  }
+}
+
+const getContext = async request => {
```
teddmason, line 44: I know it's copy and paste from other check routes, but this is not an async function.

packages/application-to-register-webapp/src/routes/developer/check-metric-file.js
```diff
@@ -0,0 +1,60 @@
+import constants from '../../utils/constants.js'
+import path from 'path'
+import { blobStorageConnector } from '@defra/bng-connectors-lib'
+
+const href = '#check-upload-correct-yes'
+const handlers = {
+  get: async (request, h) => {
+    const context = await getContext(request)
```
teddmason, line 8: no await needed, just reduce this down to `=> h.view(constants.views.DEVELOPER_CHECK_UPLOAD_METRIC, getContext(request)`

packages/application-to-register-webapp/src/routes/developer/check-metric-file.js
```diff
@@ -0,0 +1,60 @@
...
+      // delete the file from blob storage
+      const config = {
+        containerName: 'trusted',
+        blobName: metricUploadLocation
+      }
+      await blobStorageConnector.deleteBlobIfExists(config)
+      request.yar.clear(constants.redisKeys.METRIC_LOCATION)
+      return h.redirect(constants.routes.DEVELOPER_UPLOAD_METRIC)
+    } else if (checkUploadMetric === 'yes') {
+      return h.redirect('/' + constants.views.DEVELOPER_CHECK_UPLOAD_METRIC, {
```
teddmason, line 25: redirects need to be to constants.routes, no need to concatenate a / onto a view name, as we have that already in the routes array.

teddmason: Also redirects don't have a context passed

packages/application-to-register-webapp/src/routes/developer/upload-metric.js
```diff
@@ -0,0 +1,169 @@
...
+          }]
+        })
+      }
+      throw err
+  }
+}
+
+function processReturnValue (details, h) {
+  return details.resultView === constants.routes.DEVELOPER_CHECK_UPLOAD_METRIC
+    ? h.redirect(details.resultView, details.errorMessage)
```
teddmason, line 61: redirect doesn't have a context

packages/application-to-register-webapp/src/views/developer/check-metric-file.html
```diff
@@ -0,0 +1,17 @@
...
+{% set pageHeading = "Check your Biodiversity Metric 3.1 file" %}
+{% block formContent %}
+    <div class="govuk-body">
+        {% if err %}
+            {% include "errorbanner.html" %}
+        {% endif %}
+        {% set idPrefix = 'checkUploadMetric' %}
+        {% set name = 'checkUploadMetric' %}
+        {% set id = 'check-upload-correct-yes' %}
+        {% set downloadUrl = 'download-metric-file' %}
```
teddmason, line 14: I can't see a download developer metric file route added on with this pr?

teddmason (changes requested):
I know majority of comments are based on code copied from other upload routes, but they all need refactoring the same. One thing I didn't mention is that the upload route logic should use async await syntax and not promise.then() But I think this can wait for when we look to refactor all the upload routes.

Also I can't see a download developer metric file route, but the view has the value?

teddmason:
One further thing, please can you try to increase the code coverage for the PR? I think it's just a case of adding a few extra sad paths to the tests.

pwadmore-ea (changes requested):
@AjinkyaJawalkar - I agree with the changes requested from @teddmason .

Additionally, for the long term I think the metric upload routes for the landowner and developer journeys should share a common parameterised core rather than having a large amount of copy and paste between them.

I envisage the two metric upload routes being very thin and passing journey specific parameters to the common core.

I suggest that either:

- the developer journey metric upload could use the suggested common core approach now and then the landowner journey metric upload can be refactored to use the same approach at a suitable time.
- the developer journey uses the current approach temporarily and refactoring to use a common core for both metric uploads is performed as part of the wider upload refactoring mentioned by @teddmason.

PR #187, opened 2022-10-24
Feature/bngp 527 task list

reneli5:


packages/application-to-register-webapp/src/routes/land/check-legal-agreement-details.js
```diff
@@ -1,15 +1,19 @@
...
     const context = await getContext(request)
     return h.view(constants.views.LEGAL_AGREEMENT_SUMMARY, context)
   },
   post: async (request, h) => {
+    const registrationTasks = getRegistrationTasks(request)
+    registrationTasks.taskList.find(task => task.taskTitle === 'Legal information').tasks[0].status = 'COMPLETED'
+    request.yar.set(constants.redisKeys.REGISTRATION_TASK_DETAILS, registrationTasks)
     const context = await getContext(request)
-    return h.redirect(`${constants.views.LEGAL_AGREEMENT_SUMMARY}`, context)
+    return h.redirect(`/${constants.views.REGISTER_LAND_TASK_LIST}`, context)
```
teddmason, line 16: Please use the `constants.routes.REGISTER_LAND_TASK_LIST` constant as this already has the prepended slash.

Also remove the context object, it is unnecessary for a redirect.

packages/application-to-register-webapp/src/routes/land/register-land-task-list.js
```diff
@@ -1,7 +1,14 @@
 import constants from '../../utils/constants.js'
+import REGISTER_TASK_LIST from '../../utils/register-task-list.js'
 
 const handlers = {
-  get: async (_request, h) => h.view(constants.views.REGISTER_LAND_TASK_LIST)
+  get: async (request, h) => {
+    let dataContent = request.yar.get(constants.redisKeys.REGISTRATION_TASK_DETAILS)
+    if (dataContent === null || dataContent === undefined) {
```
teddmason, line 7: Couldn't the null and undefined checks be replaced with a falsey statement. ie `if (!dataContent)`

And then could it not be reduced to `const dataContent = request.yar.get(constants.redisKeys.REGISTRATION_TASK_DETAILS) || REGISTER_TASK_LIST`

packages/application-to-register-webapp/src/utils/constants.js
```diff
@@ -101,8 +101,11 @@ const LANDOWNERS = 'landowners'
 const LANDOWNER_CONSENT = 'land/landowner-consent'
 const LANDOWNER_CONSENT_KEY = 'landowner-consent'
 const CHECK_OWNERSHIP_DETAILS = 'land/check-ownership-details'
+const REGISTRATION_TASK_DETAILS = 'registrationTaskDetails'
+const DEFAULT_REGISTRATION_TASK_STATUS = 'NOT STARTED'
 const REFERRAL_PAGE_LIST = [
-  `/${LEGAL_AGREEMENT_SUMMARY}`
+  `/${LEGAL_AGREEMENT_SUMMARY}`,
+  REGISTRATION_TASK_DETAILS
```
teddmason, line 108: I'm not sure the need to add REGISTRATION_TASK_DETAILS here, as it shouldn't be caught at all by the referer logic.  And also please note that with BNGP-190 this object no longer exists for the referer routing

packages/application-to-register-webapp/src/utils/helpers.js
```diff
@@ -57,11 +59,20 @@ const getReferrer = (request, referrerId, unsetFlag) => {
   return currentReferrer
 }
 
+const getRegistrationTasks = request => {
```
teddmason, line 62: Why does the whole request object need to be passed to the function, surely only the value for registrationTasks?

packages/application-to-register-webapp/src/utils/helpers.js
```diff
@@ -57,11 +59,20 @@ const getReferrer = (request, referrerId, unsetFlag) => {
...
+  const registrationTasks = request.yar.get(constants.redisKeys.REGISTRATION_TASK_DETAILS)
+  if (registrationTasks === undefined || registrationTasks === null) {
```
teddmason, line 64: again does `!registationTasks` not cover this?

packages/application-to-register-webapp/src/views/land/legal-agreement-type.html
```diff
@@ -32,7 +32,7 @@
                 {
                     value: "Conservation covenant",
                     text: "Conservation covenant",
-                    id: "conversation-covenant",
+                    id: "conversation",
```
teddmason, line 35: spelling

packages/application-to-register-webapp/src/routes/land/check-legal-agreement-details.js
reneli5: Removed the context object. But Left the constant as it is the same in master branch

packages/application-to-register-webapp/src/utils/constants.js
reneli5: Can we have a discussion on the referrer routing. I add it here to have a list of pages that can be considered as referral candidates. Not all pages can be referred from.

packages/application-to-register-webapp/src/utils/helpers.js
reneli5: We are passing the whole request so that we can get yar object in one location and not duplicate the code

packages/application-to-register-webapp/src/routes/land/register-land-task-list.js
reneli5: Use a falsey

packages/application-to-register-webapp/src/utils/constants.js
teddmason: See here https://github.com/DEFRA/biodiversity-net-gain-service/blob/82b2a7572f26936d13ed9e7d559f2e41b21b1e20/packages/application-to-register-webapp/src/utils/constants.js#L267

reneli5: Same concept. I agree that we don't need REGISTRATION_TASK_DETAILS.
Is your checkRoutes array merged to master?

packages/application-to-register-webapp/src/utils/helpers.js
teddmason: Why has this been resolved? I can't see it fixed in the code

packages/application-to-register-webapp/src/routes/land/check-legal-agreement-details.js
```diff
@@ -1,15 +1,16 @@
...
 const handlers = {
   get: async (request, h) => {
     const context = await getContext(request)
     return h.view(constants.views.LEGAL_AGREEMENT_SUMMARY, context)
   },
   post: async (request, h) => {
-    const context = await getContext(request)
-    return h.redirect(`${constants.views.LEGAL_AGREEMENT_SUMMARY}`, context)
+    processCompletedRegistrationTask(request, 'Legal information')
+    return h.redirect(`/${constants.views.REGISTER_LAND_TASK_LIST}`)
```
teddmason, line 13: switch to use a constants.routes value

teddmason (changes requested):

packages/application-to-register-webapp/src/utils/constants.js
reneli5: You are describing what I have implemented. Have you looked at the code?

packages/application-to-register-webapp/src/utils/helpers.js
reneli5: It needs the request. Please look at the code.

packages/application-to-register-webapp/src/routes/land/check-legal-agreement-details.js
reneli5: Switched.

packages/application-to-register-webapp/src/routes/land/check-legal-agreement-details.js
reneli5: Updated

packages/application-to-register-webapp/src/utils/helpers.js
teddmason: No there is not, but it is basic

packages/application-to-register-webapp/src/utils/helpers.js
teddmason: I have looked at the code as a whole and I disagree, functional it makes no difference so be it.

packages/application-to-register-webapp/src/utils/register-task-list.js
```diff
@@ -0,0 +1,63 @@
+import constants from './constants.js'
+
+const REGISTER_TASK_LIST = {
```
teddmason, line 3: One last thing, just realised. This object shouldn't be in capitalised snake case, as it is not a constant as such, but an object/class that is manipulated.  Please change to pascal, and all usage of it in code files.

teddmason (changes requested):

packages/application-to-register-webapp/src/utils/register-task-list.js
reneli5: It is a constant and the code does not manipulate that object directly as such. Every time the code makes a copy of it, leaving it intact.

reneli5: Renamed to camelcase

teddmason (approved):

PR #194, opened 2022-11-02
Feature/bngp 1267 accessibility

reneli5:


packages/application-to-register-webapp/src/views/land/add-legal-agreement-parties.html
```diff
@@ -83,7 +83,6 @@ <h1 class="govuk-heading-l" tabindex="-1">
                                     name: "organisation[" + i + "][role]",
                                     value: "County Council",
                                     text: "County Council",
```
teddmason, line 85: To be consistent with the ID updates on other inputs this ID should be updated to align with county council

packages/application-to-register-webapp/src/views/land/add-legal-agreement-parties.html
```diff
@@ -83,7 +83,6 @@ <h1 class="govuk-heading-l" tabindex="-1">
...
-                                    label: "County Council",
                                     checked: roleRowIndex.county_council,
                                     classes: "govuk-input--width-2",
                                     attributes: {
```
teddmason, line 88: The inputs data-name and data-id attributes aren't set correctly, think they need the same concatenation as the id/name fields. As it comes through to the HTML as "%index%" . I realise they're not part of the bug, but probably should be fixed

teddmason (changes requested):

pwadmore-ea (changes requested):
@reneli5 - I've scanned this page with the Axe browser extension following the revisions requested by @teddmason.

I can see one accessibility issue remaining that is page specifc:

image

pwadmore-ea (approved):
Having spoken to @shaundefra  I understand that the page specific accessibility error is present in the HTML prototype. I'm happy to approve this pull request accordingly and will add a comment to the story as a reminder for resolution when the prototype is corrected.

pwadmore-ea (dismissed):

teddmason (approved):

PR #205, opened 2022-11-21
Feature/bngp 1410

teddmason:


pwadmore-ea (commented):
I have been unable to submit a journey as a single landowner due to failing landowner consent validation. Is the landowner consent indicator being set to false in this scenario?

teddmason:
> I have been unable to submit a journey as a single landowner due to failing landowner consent validation. Is the landowner consent indicator being set to false in this scenario?

Should be fixed now. Gave landownerConsent a default of false, and the otherLandowners array a default of an empty array

pwadmore-ea (approved):

PR #221, opened 2023-01-26
BNGP-1804 Upgrade SignalR emulator for net6.0 use

pwadmore-ea:
https://eaflood.atlassian.net/browse/BNGP-1804


teddmason (dismissed):
Re built my docker images and re run unit tests, and did a quick run through of signalr connectivity on local app via file upload and response.

PR #234, opened 2023-02-21
Updated changes as per LO implementation of metric extraction

AjinkyaJawalkar:


packages/webapp/src/utils/get-build-config.js
```diff
@@ -0,0 +1,50 @@
+import { handleEvents } from './azure-signalr.js'
```
teddmason, line 1: Good idea to extract the upload config for the metric file, but I think it would be better to make it file type agnostic. As METRIC_UPLOAD_TYPE is hard coded, it would be better to make getBuildConfig to accept a file upload type so that it can be reused elsewhere in the application, especially as the name of the file is file type agnostic

pwadmore-ea (changes requested):
I think this represents good progression towards common metric extraction and rendering functionality for the landowner and developer journeys.

I hope my comments (those on increased commonality in particular) make sense.

PR #235, opened 2023-02-22
Feature/bngp 523 metric playback

teddmason:


pwadmore-ea (changes requested):
@teddmason - I am happy with the changes made for metric data playback. However, while running the application with these changes I think I have found a little bug in the extraction configuration.

When viewing the habitat creation and enhancement data from an uploaded Biodiversity Metric 4.0 document, it seems like the original river lengths within the F3 tab are being displayed rather than proposed values associated with the creation/enhancement.

If I am correct I think https://github.com/DEFRA/biodiversity-net-gain-service/blob/master/packages/bng-metric-service/src/helpers/extractors/extractionconfig/metric/v4.0/f3OffSiteWaterCEnhancement.js needs a line adding to source the river length values from the __EMPTY_1 attribute (as done for other tabs).

I've added this line manually locally, refactored the affected unit test and it seems to be displaying the correct information locally (assuming the bug is genuine).

__EMPTY_1: 'Length (km)',

pwadmore-ea (approved):

pwadmore-ea (approved):

PR #238, opened 2023-02-24
Feature/BNGP-1851 - Add Minimum Date Validation

z424brave:
https://eaflood.atlassian.net/browse/BNGP-1851

Add validation to habitat-works-start-date, legal-agreement-start-date and management-monitoring-start-date to chack that the dates entered are greater than or equal to 30th January 2020.


packages/webapp/src/utils/helpers.js
```diff
@@ -3,10 +3,22 @@ import constants from './constants.js'
 import registerTaskList from './register-task-list.js'
 import validator from 'email-validator'
 
-const validateDate = (payload, ID, desc) => {
-  const day = payload[`${ID}-day`]
-  const month = payload[`${ID}-month`]
+const isoDateFormat = 'YYYY-MM-DD'
+
+const parsePayload = (payload, ID) => {
+  const day = (payload[`${ID}-day`] && payload[`${ID}-day`].length === 1) ? payload[`${ID}-day`].padStart(2, 0) : payload[`${ID}-day`]
```
teddmason, line 9: Should `.padStart` pad with a string? Probably makes no difference

z424brave: I agree it will probably make no difference but I will make it explicitly a string for clarity

pwadmore-ea (approved):

PR #241, opened 2023-03-01
Feature/BNGP-1993 Implement Google Tag Manager

z424brave:
https://eaflood.atlassian.net/browse/BNGP-1993

Include code snippet for Google Tag Manager in layout.html so it is included in all pages for the Land owner journey.


docker/azure-services/local-webapp-env-template
```diff
@@ -14,3 +14,4 @@ export MAX_GEOSPATIAL_LAND_BOUNDARY_UPLOAD_MB="50"
 # Change to http://azure_services:7071/api if the application to register web application is running in a container.
 export SIGNALR_URL="http://azure_services:8082/api"
 export MAX_METRIC_UPLOAD_MB="50"
+export GOOGLE_TAGMANAGER_ID="GTM-PL85H6J"
```
teddmason, line 17: For local development purposes app should accept a blank tagmanager ID and skip its loading, so as not to send data to GTM

packages/webapp/client/js/tagmanager_datalayer.js
teddmason, line 1: File name needs to be kebab case

packages/webapp/client/js/tagmanager_head.js
teddmason, line 1: kebab case again

packages/webapp/webpack.config.js
```diff
@@ -33,5 +36,10 @@ export default {
         }
       }
     ]
-  }
+  },
+  plugins: [
+    new webpack.EnvironmentPlugin({
+      GOOGLE_TAGMANAGER_ID: 'GTM-PL85H6J' // use 'GTM-PL85H6J' unless process.env.GOOGLE_TAGMANAGER_ID is defined
```
teddmason, line 42: This needs to be pulled from the environment variable, not hardcoded

packages/webapp/client/js/core.js
```diff
@@ -14,3 +27,80 @@ const jsElements = document.getElementsByClassName('defra-js-show')
...
+  const prefs = {
+    analytics: accepted ? 'on' : 'off'
+  }
+  window.bng.utils.setCookie(cookieUserPreference, JSON.stringify(prefs), cookieUserPreferenceExpiry)
+  cookiePageBanner.style.display = 'block'
+  cookiePageBanner.setAttribute('tabindex', '-1')
+  cookiePageBanner.focus()
+}
+
+if (cookiePrefsPage) {
```
teddmason, line 55: Can cookiesPrefsPage code not be moved to its own page script file? Was looking for  `cookies.js`

packages/webapp/client/js/core.js
```diff
@@ -14,3 +27,80 @@ const jsElements = document.getElementsByClassName('defra-js-show')
...
+  const saveButton = document.querySelector('#cookies-save')
+  cookiePageBanner.style.display = 'none'
+  saveButton?.addEventListener('click', function (event) {
```
teddmason, line 58: Also when testing if setting analytics preferences on the cookies page it isn't setting the banner seen cookie, so banner still shows after setting cookies, needs to be hidden I would have thought

packages/webapp/client/js/core.js
```diff
@@ -14,3 +27,80 @@ const jsElements = document.getElementsByClassName('defra-js-show')
...
+const cookieSeenBannerExpiry = 30
+const cookieUserPreference = 'set_cookie_usage'
+const cookieUserPreferenceExpiry = 30
+
+const savePreference = accepted => {
+  const prefs = {
+    analytics: accepted ? 'on' : 'off'
+  }
+  window.bng.utils.setCookie(cookieUserPreference, JSON.stringify(prefs), cookieUserPreferenceExpiry)
+  cookiePageBanner.style.display = 'block'
```
teddmason, line 50: Accepting or rejecting on banner is throwing a JS error on setting this style.display, cookiePageBanner is null

teddmason: Presumably it only needs to execute if on cookies page, giving more weight to needing a cookies.js script file

packages/webapp/src/views/cookie-banner.html
```diff
@@ -0,0 +1,82 @@
+{% from "govuk/components/cookie-banner/macro.njk" import govukCookieBanner %}
+<noscript>
+  {{ govukCookieBanner({
+  ariaLabel: serviceName,
+  messages: [
+    {
+      headingText: "Cookes on " + serviceName,
```
teddmason, line 7: Spelling of cookies

packages/webapp/src/views/cookie-banner.html
```diff
@@ -0,0 +1,82 @@
...
+      text: "We use cookies to make this service work and collect analytics information. To accept or reject cookies, turn on JavaScript in your browser settings or reload this page."
+    }
+  ]
+}) }}
+</noscript>
+<div class="js-cookie-banner-container">
```
teddmason, line 13: Banner is still showing if no scripting is enabled

packages/webapp/client/js/core.js
```diff
@@ -14,3 +27,80 @@ const jsElements = document.getElementsByClassName('defra-js-show')
...
+    savePreference(true)
+  })
+
+  rejectButton?.addEventListener('click', function (event) {
+    showBanner(rejectedBanner)
+    event.preventDefault()
+    savePreference(false)
+  })
+
+  acceptedBanner?.querySelector('.js-hide').addEventListener('click', function () {
```
teddmason, line 97: So it looks as though the banner only hides itself if user actively uses "Hide Banner" after choosing cookies choice on banner.  This cookie needs to be set after they've chosen to accept or deny tracking imo.  Because even if not atm if you chose tracking but don't hide the banner, the next page shows a banner asking to accept/deny tracking. This is linked to my comment saying if accepting tracking on cookies page the banner should hide as well.

teddmason (changes requested):
Fundamental issue is that this code is loading the GTM snippet and broadcasting to google on every page view before a user has even accepted tracking. And therefore if they reject tracking then something is still being broadcast to GTM.  I'm aware that performance analyst asking for the datalayer to set tracking to no or yes. but code wise we need to disable the GTM snippet running until a user as accepted tracking

teddmason:
For comparison purposes take a look at gov.uk and install google tag assistant Chrome plugin and record. You can see that they do no tracking at all on first load, if reject there is no tracking, it only tracks on accept. Whilst also then hiding the banner on follow up pages

packages/webapp/client/js/core.js
z424brave: Currently left all code in core.js  - rather than seperate cookies page only into a seperate file.

pwadmore-ea (approved):
The first two items from this feedback given outside GitHub initially have been resolved.

- noscript cookie banner does not include a View Cookies link (unlike existing services)
- inconsistency between cookie title when the service is used with and without JavaScript
  - Cookies on Register land as a biodiversity gain site (noscript)
  - Register land as a biodiversity gain site (JavaScript)
- Cookies page radio buttons do not default to the current acceptance state.

As the remaining item is not part of the current acceptance criteria it is proposed that resolution is addressed in a subsequent story.

pwadmore-ea (dismissed):

pwadmore-ea (dismissed):

pwadmore-ea (approved):
Reapproving following the addition of functionality to:

- show analytics cookie acceptance state on the cookies page
- remove analytics cookie when they are rejected.

I note that the correctness of the cookies listed in the cookies page versus what is actually used at runtime will be addressed in a subsequent story.

teddmason:
All looks fine to me, apart from I feel it needs a wrapper around the GTM loading script to ensure that the GTM ID is set before executing, otherwise returns a console error because GTM ID is blank

pwadmore-ea (approved):

teddmason:
fine for me now

pwadmore-ea (approved):

teddmason (approved):

PR #242, opened 2023-03-08
Added new page to get full name

usman-j-shaikh:


pwadmore-ea (commented):
I have observed that names entered through the /land/name and /developer/details-name routes cause the same name to be auto-populated when the other route is accessed.

Rather than simply including the name view from the landowner journey I'm wondering if something like a parametrised partial is needed to help distinguish between data used by the two journeys.

@teddmason  - I'd appreciate your thoughts on this.

teddmason:
> I have observed that names entered through the /land/name and /developer/details-name routes cause the same name to be auto-populated when the other route is accessed.
> 
> Rather than simply including the name view from the landowner journey I'm wondering if something like a parametrised partial is needed to help distinguish between data used by the two journeys.
> 
> @teddmason - I'd appreciate your thoughts on this.

I was surprised to see the whole view being reused and not being pulled in as a partial, but assumed if it had no unexpected consequences it would be fine, but if you found that then I guess it needs moving to a partial

usman-j-shaikh:
> I have observed that names entered through the /land/name and /developer/details-name routes cause the same name to be auto-populated when the other route is accessed.
> 
> Rather than simply including the name view from the landowner journey I'm wondering if something like a parametrised partial is needed to help distinguish between data used by the two journeys.
> 
> @teddmason - I'd appreciate your thoughts on this.

I am using routes as per prototype of Developer Journey

teddmason (approved):

PR #244, opened 2023-03-10
feature/BNGP-2174: Additional Applicant details - Email address

usman-j-shaikh:
To add an additional field within the applicant detail:
Email address


packages/webapp/src/views/developer/details-email.html
```diff
@@ -0,0 +1,3 @@
+{% set serviceName = "Record biodiversity gains" %}
+
+{% include 'land/email.html' %}
```
teddmason, line 3: I think you're going to have the same problem with the email as the name view, needs to be its own view or both landowner and developer journey views need to pull in a partial

teddmason (changes requested):
I think you're going to have the same problem with the email as the name view, needs to be its own view or both landowner and developer journey views need to pull in a partial

teddmason (approved):

PR #246, opened 2023-03-14
Feature/bngp 2175 email address correct

usman-j-shaikh:
This PR includes changes about email address confirmation.
- Email address validator
- Confirm email address page for DJ
- Modified helper method

### DIFF
--- a/packages/webapp/src/utils/constants.js
+++ b/packages/webapp/src/utils/constants.js
@@ -171,6 +171,7 @@ const DEVELOPER_FULL_NAME = 'developer-fullname'
 const DEVELOPER_REFERER = 'developer-referer'
 const DEVELOPER_DETAILS_EMAIL = 'developer/details-email'
 const DEVELOPER_EMAIL_VALUE = 'developer-email-value'
+const DETAILS_EMAIL_CONFIRM = 'developer/details-email-confirm'
 
 const confirmFileUploadOptions = {
   NO,
@@ -407,7 +408,8 @@ const routes = {
   CHECK_HABITAT_CREATED,
   CHECK_METRIC_DETAILS,
   DEVELOPER_DETAILS_NAME,
-  DEVELOPER_DETAILS_EMAIL
+  DEVELOPER_DETAILS_EMAIL,
+  DETAILS_EMAIL_CONFIRM
 }
 
 const uploadErrors = {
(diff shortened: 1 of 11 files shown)


packages/webapp/src/views/developer/details-email-confirm.html
```diff
@@ -0,0 +1,86 @@
...
+
+{% block formContent %}
+<div class="govuk-body">
+    {% if err %}
+        {% include "errorbanner.html" %}
+    {% endif %}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            {% set emailHtml %}
+                {% if errorMessage %}
```
teddmason, line 14: Don't think it needs the if statement here just to set the errorMessage, errorMessage could just be set as an empty object/array if not in error

packages/webapp/src/views/developer-form-layout.html
teddmason, line 1: How come you need this file? looks to be nearly the same as the current form-layout.html file?

packages/webapp/src/utils/constants.js
```diff
@@ -171,6 +171,8 @@ const DEVELOPER_FULL_NAME = 'developer-fullname'
 const DEVELOPER_REFERER = 'developer-referer'
 const DEVELOPER_DETAILS_EMAIL = 'developer/details-email'
 const DEVELOPER_EMAIL_VALUE = 'developer-email-value'
+const DEVELOPER_DETAILS_EMAIL_CONFIRM = 'developer/details-email-confirm'
+const DEVELOPER_CONFIRM_EMAIL = 'developer-yes'
```
teddmason, line 175: This redis key needs a bit more description probably something like developer-confirm-email, developer-yes doesn't sound right?

teddmason (changes requested):

packages/webapp/src/views/developer-form-layout.html
usman-j-shaikh: We have different page title, and to define this title each and every page not sound good
And I think more changes will come in future which will slightly differ than LOJ
So I want to centralize this work using this file

pwadmore-ea (changes requested):

pwadmore-ea (approved):

teddmason (approved):

PR #266, opened 2023-04-12
Feature/bngp 2193 eligibility results updated

AjinkyaJawalkar:


z424brave (dismissed):
Looks good to me

packages/webapp/src/utils/developer-constants.js
```diff
@@ -43,6 +43,11 @@ const ROUTING_REGISTER_OPTIONS = {
   RECORD: 'record-off-site'
 }
 
+const developerEligibilityHTML = {
```
usman-j-shaikh, line 46: @AjinkyaJawalkar 
Is it ok to move this logic in route, instead of defining constant?
cc @z424brave @pwadmore-ea

usman-j-shaikh (commented):
There are some comment as per my understanding, so if its not needed, then please ignore.

usman-j-shaikh (approved):

pwadmore-ea (changes requested):

packages/webapp/src/utils/developer-constants.js
AjinkyaJawalkar: Marking this comment as resolved as per discussed over call

PR #267, opened 2023-04-12
fix/BNGP-2572-geospatial-upload-regression

pwadmore-ea:
https://eaflood.atlassian.net/browse/BNGP-2572

Copy NTv2 format files at build time due to read only file systems in the official cloud environments. Use a postinstall script that does not assume a particular directory structure when copying.


AjinkyaJawalkar (approved):
Looks good to me

packages/geoprocessing-service/src/helpers/__tests__/copy-ntv2-format-files.spec.js
```diff
@@ -1,7 +1,5 @@
 import fs from 'fs'
 import path from 'path'
-import copyNtv2FormatFiles from '../copy-ntv2-format-files.cjs'
-
 console.log(process.cwd() + '**')
```
usman-j-shaikh, line 3: @pwadmore-ea 
Is this console necessary now though?

usman-j-shaikh, line 3: @pwadmore-ea 
Is this console necessary now though?

packages/geoprocessing-service/package.json
```diff
@@ -1,11 +1,15 @@
 {
   "name": "@defra/bng-geoprocessing-service",
-  "version": "0.75.0",
+  "version": "0.75.1",
   "type": "module",
   "private": false,
   "description": "Geoprocessing service for Biodiversity Net Gain",
-  "main": "src/service.js",
```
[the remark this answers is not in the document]
pwadmore-ea: @usman-j-shaikh - The main field in package.json has been succeeded by the exports field in newer versions of Node.js (please see https://nodejs.org/api/packages.html#package-entry-points). As this patch relies on the package having multiple entry points I've had to switch from using main to exports. As part of continuous improvement we could consider refactoring all packages to use exports but I do not see this as a current priority.

packages/geoprocessing-service/src/helpers/__tests__/copy-ntv2-format-files.spec.js
pwadmore-ea: Refactored

usman-j-shaikh (approved):

PR #280, opened 2023-05-04
Feature/bngp 1922

teddmason:
- Added in pre option function for routes that need to check for applicant details in session
- Added method to inject session data prior to using server.inject, so that we can manipulate the session within the unit test
- Updated a load of route tests to use the standard server.inject as they are now able to complete correctly due to the presence of the session data

### DIFF
--- a/packages/webapp/src/__mocks__/on-pre-handler.js
+++ b/packages/webapp/src/__mocks__/on-pre-handler.js
@@ -1,0 +1,18 @@
+import application from '../../src/__mock-data__/test-application.js'
+
+// Adds prehandler to server that loads session with mock application data or sessionData object provided from test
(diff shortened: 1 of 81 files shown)


pwadmore-ea (changes requested):
Just a couple of small comments from me that are both linked to which routes should have the applicant details check included:

- It looks like the check needs adding to the landowner journey download routes at minimum. If I try and access one of the download routes in a new session currently, the generic error page is displayed.
  - The split between which other landowner journey routes do and do not perform the check seems sensible to me.
- Should the check be added to developer journey routes as well as part of this story or should it part be part of a distinct story?

pwadmore-ea (approved):
Approved following changes

PR #282, opened 2023-05-08
Feature/BNGP-2680: Add name to additional email addresses

usman-j-shaikh:
Includes changes for existing page of additional email address
- Optimized code
- Added duplicate email address check
- Added script to validate number of 1 fields rows
- Improved test coverage

### DIFF
--- a/packages/webapp/src/routes/developer/email-entry.js
+++ b/packages/webapp/src/routes/developer/email-entry.js
@@ -1,5 +1,5 @@
 import constants from '../../utils/constants.js'
-import { emailValidator, getErrById } from '../../utils/helpers.js'
+import { emailValidator, getErrById, validateName } from '../../utils/helpers.js'
 
 const handlers = {
   get: async (request, h) => {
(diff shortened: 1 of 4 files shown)


packages/webapp/src/routes/developer/email-entry.js
```diff
@@ -20,19 +20,49 @@ const handlers = {
...
-  for (const i in payload) {
-    const email = payload[i]
-    const result = emailValidator(email, `#emailAddresses[${i}]`)
-    if (result) {
-      err.push(result.err[0])
+  const fullNames = request.payload.fullNames
+  const emails = request.payload.emails
+
+  if (Array.isArray(emails)) {
+    for (const i in emails) {
```
teddmason, line 28: I feel like this loop should be in the processFieldValues, but not necessarily a must

AjinkyaJawalkar (approved):

packages/webapp/src/routes/developer/email-entry.js
usman-j-shaikh: I appreciate your suggestion. In my opinion, the same block code is used in both conditions, so I have added this method outside of the loop. Still, I'm open to your advice.
Thanks!

teddmason (approved):

AjinkyaJawalkar:
Looks good to me!

PR #289, opened 2023-05-19
Tech debt/BNGP-2857: common upload code for developer upload metric 

usman-j-shaikh:
It includes following changes:

- Added common upload code for developer metric file upload
- Azure function modified to implement validation
- Unit test coverage as per newly added code


packages/webapp/src/utils/helpers.js
```diff
@@ -445,6 +445,25 @@ const getHumanReadableFileSize = (fileSizeInBytes, maximumDecimalPlaces = 2) =>
   return `${parseFloat(humanReadableFileSize.toFixed(parseInt(maximumDecimalPlaces)))} ${units}`
 }
 
+const getValidation = (metricValidation, href) => {
```
teddmason, line 448: Can you change the function name to be more verbose/descriptive, getValidation is too generic (as I know it was pulled straight from the uploadMetric handler where it would be fine, but in the helpers object it needs to be more descriptive

teddmason (changes requested):

teddmason (approved):

teddmason (approved):

PR #291, opened 2023-05-22
Fix metric upload bug

teddmason:
As part of testing AV scanning there is a bug where by if a metric successfully gets AV scanned and returned then the app is throwing a "blob not found error" and this is due to the assumption that the function message will have the containerName set, however when the message comes from AV scanning service bus this is not the case. So fix is to set the containerName manually.


paulsimonandrews (approved):
LGTM

pwadmore-ea (approved):
The upload of a non-Biodiversity Metric spreadsheet uses generic file upload error handling rather than user friendly context sensitive error handling.

A follow on story to provide user friendly context sensitive error handling is suggested.

teddmason:
Closing this off as have recreated the branch based on master as running rebase was a nightmare.  Created a branch from master then merged in this branch and stripped out any debug/testing code not needed in prod.

teddmason:
New pr https://github.com/DEFRA/biodiversity-net-gain-service/pull/370

PR #296, opened 2023-05-26
Feature/bngp 2213 application summary

AjinkyaJawalkar:
It contains UI of check-answers page


packages/webapp/src/routes/__tests__/developer/check-answers.spec.js
```diff
@@ -1,12 +1,156 @@
...
 
 describe(url, () => {
   describe('GET', () => {
-    it(`should render the ${url.substring(1)} view`, async () => {
-      const response = await submitGetRequest({ url })
-      expect(response.statusCode).toBe(200)
+    it(`should render the ${url.substring(1)} view`, done => {
+      jest.isolateModules(async () => {
+        try {
+          const session = setDeveloperApplicationSession()
```
teddmason, line 12: btw submitGetRequest now automatically populates with session data so that routes needing session data don't fail on server.inject, you can add additional fields to the session object for the developer journey if necessary
see webapp/src/__mocks__/on-pre-handler.js

teddmason (commented):
How can this pr be tested if the corresponding function in the functionapp doesn't exist for process developer application?

pwadmore-ea (changes requested):
- The metric file added is showing as null on the application summary page.
- All of the change links for the metric are going back to the upload metric page rather than the pages in the ACs/prototype
bngp-2213

pwadmore-ea (dismissed):

usman-j-shaikh (approved):

PR #314, opened 2023-05-31
Enable github app to commit versioning to protected branches

teddmason:


paulsimonandrews (approved):
LGTM.

I've seen https://github.com/getsentry/action-github-app-token used where you have tibdex/github-app-token, but they seem to do exactly the same thing.

teddmason:
> LGTM.
> 
> I've seen https://github.com/getsentry/action-github-app-token used where you have tibdex/github-app-token, but they seem to do exactly the same thing.

Yes I've seen people using both, 50:50 in the end as they both do the same thing.

PR #347, opened 2023-06-26
Attribute id and test-id added to all continue buttons in DJ

AjinkyaJawalkar:
Attribute id and test-id added to all continue buttons in DJ


shaundefra:
Tested locally on the dev journey ac tests

PR #349, opened 2023-06-27
Feature/bngp 2867 defra

teddmason:


teddmason:
Switching base branch to the overall defra ID feature branch feature/defra-id-complete

lerna.json
```diff
@@ -2,5 +2,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.136.0"
+  "version": "0.136.2"
```
paulsimonandrews, line 5: Isn't this done via the Github action?

teddmason: I know you messaged on slack, but just to answer here by running lerna publish we can publish our feature + packages to npm. It tags the codebase thereby triggering our azure artefact builds and we can deploy a feature branch to an azure environment.

When the feature is ultimately merged into master all the versioning stuff will be overwritten

paulsimonandrews (approved):
LGTM.

There's a console.log in a test you might want to remove, but no need to hold up the approval

packages/webapp/src/utils/__tests__/constants.spec.js
```diff
@@ -0,0 +1,28 @@
+describe('Constants', () => {
+  beforeEach(() => {
+    jest.resetModules()
+  })
+  it('Should load test routes if NODE_ENV is set to development', () => {
```
paulsimonandrews, line 5: Typo: "if NODE_ENV is set to development" should be test not development

PR #359, opened 2023-07-02
BNGP 2800 Rounding Error Bug

paulsimonandrews:
Fix the bug reported in ticket https://eaflood.atlassian.net/browse/BNGP-2800

Floating point rounding error when checking total values from excel metric file. Add a tolerance to the check to cope with floating point rounding errors.


teddmason (changes requested):
1 code smell to sort from sonarcloud

teddmason (approved):

PR #366, opened 2023-07-04
feat: BNGP-3125 legal agreement content changes

KiranVarma02:
https://eaflood.atlassian.net/browse/BNGP-3125


packages/webapp/src/routes/land/add-legal-agreement-parties.js
```diff
@@ -158,19 +162,25 @@ const handlers = {
       inProgressUrl: constants.routes.ADD_LEGAL_AGREEMENT_PARTIES
     })
     const partySelectionData = request.yar.get(constants.redisKeys.LEGAL_AGREEMENT_PARTIES)
+    const documentType = request.yar.get(constants.redisKeys.LEGAL_AGREEMENT_DOCUMENT_TYPE)?.toLowerCase()
```
teddmason, line 165: Would it be better to get the legalAgreementType here? So as not to duplicate the getLegalAgreementDocumentType(documentType) in the return h.view...

Only a could

teddmason (commented):
One tiny thing

packages/webapp/src/routes/land/add-legal-agreement-parties.js
KiranVarma02: agree. I have applied same change for all files.

teddmason (approved):

teddmason (approved):

teddmason (approved):

PR #370, opened 2023-07-06
Feature/bngp 2850 latest

teddmason:
This includes bug fix https://eaflood.atlassian.net/browse/BNGP-2912

As well as some retry functionality.


packages/azure-functions/ProcessUntrustedFile/index.mjs
```diff
@@ -54,7 +57,11 @@ export default async function (context, message) {
     let signalRMessageArguments
     if (err instanceof ThreatScreeningError) {
       signalRMessageArguments = [{ threatScreeningDetails: err.threatScreeningDetails }]
+    } else if (context?.bindingData?.dequeueCount < maximumNumberOfReplayAttempts) {
+      // Replay the message
+      throw err
```
KiranVarma02, line 62: Just for my understanding, What is the plan here please? until we add some logic for replay we want to throw err?

teddmason: So this is piggybacking the function app replay logic.  If an error is thrown the function app will attempt a full replay. Using the dequeue count we can figure out how many attempts have been made. Finally on the 5th attempt or the max count we just return a signalr message, so function app thinks the run was successfully, although in reality it might have failed. But we return the failed message to front end

KiranVarma02 (approved):
LGTM

PR #387, opened 2023-07-17
feature/BNGP-3310-title-change

AjinkyaJawalkar:
Added new condition in <layout.html> to show page title conditionally depends on Error on page and passed respective values from all the views


paulsimonandrews (changes requested):
Try just updating layout.html to access `err` directly:

{% block pageTitle %}
{% if err %} Error: {% endif %} {{ pageHeading }} {{ titleSuffix }}
{% endblock %}

Then you don't need to update all of the other templates.

PR #391, opened 2023-07-21
Feature/BNGP-3036: Update Landowner metric file extraction to include additional fields required for allocation

usman-j-shaikh:


pwadmore-ea (changes requested):
@usman-j-shaikh  - These changes are causing a HTTP 500 response error on the web application when attempting to process an uploaded landowner metric. @shaundefra has confirmed the same problem.

Please investigate.

pwadmore-ea (approved):

PR #393, opened 2023-08-04
Feature/BNGP-3402 and 3404 disabling routes

usman-j-shaikh:
Added filter to the routes
Added a helper method to check env
Conditionally redirected and displayed routes


pwadmore-ea (dismissed):
@usman-j-shaikh / @paulsimonandrews  - Having reviewed this pull request in detail I believe the current approach to disabling routes using a delimited environment variable is too brittle and could lead to risk of features being enabled accidentally. For example, a typo in or the omission of a route to be disabled in the environment variable specifying disabled routes could cause intended consequences. Suitable acceptance tests are needed to catch these types of errors.

I think it is going to be safer to create a simple environment variable based feature flag for each affected area of functionality and use those to control availability of the associated routes. For example:

- ENABLE_GEOSPATIAL_SUPPORT_FOR_REGISTRATIONS
- ENABLE_ADDITIONAL_EMAIL_ADDRESSES_FOR_ALLOCATIONS
  
I think using environment variables to enable functionality rather than disable it is more straightforward, requires no effort in the cloud environment configuration to disable functionality for MVP and mitigates risk of incorrect environment variable based route configuration accordingly.

Internally, most of the existing logic associated with disabled routes can remain as it is but logic will need to be created to:

- Associate each feature flag with one or more routes.
- Ensure the route(s) associated with each feature flag are enabled/disabled correctly.

I think this approach will be safer and easier to test accordingly.

packages/webapp/src/utils/constants.js
```diff
@@ -492,6 +492,14 @@ const minStartDates = {
   MANAGEMENT_MONITORING_MIN_START_DATE
 }
 
+const ROUTES_TO_BE_DISABLED = {
```
teddmason, line 495: I think this is unnecessary, the code is already hardcoding which routes can be disabled in the helpers file when pulled through

teddmason (changes requested):
Hi @usman-j-shaikh I think this work is overcomplex for what needs to be going on here. Where this code has key functionality in constants, helpers and router. I think it can easily be boiled down to only being in the router file. I think it would be a lot easier to maintain this functionality going forward by keeping it simple and in the router file.

Just some simple If statements would be best for this, no complex loops and imports, nothing overly complex.

eg (i'm just making this up from my head)
`import { routes } from 'constants.js'`
`if (process.env.ENABLE_ROUTE_SUPPORT_FOR_GEOSPATIAL) {
  routes.filter(routes.CHECK_GEOSPATIAL_FILE) // filter is more complex than this obviously
  ...
}
if (process.env.ENABLE.....)
...
`

I spoke to @pwadmore-ea as well and agreed.  What you've done here looks great, but I'd like to advocate "keep it simple" principals.

teddmason (dismissed):
Thank you for refactoring, the code feels a lot more cleaner and simplified now

teddmason (dismissed):

teddmason (approved):

shaundefra:
tested by @LeenaKu

PR #400, opened 2023-08-14
Credits Estimator BNGP-3167 BNGP-2892 BNGP-2888

paulsimonandrews:
Implement Credits Estimator


packages/webapp/src/credits/calculate.js
```diff
@@ -0,0 +1,18 @@
+import unitPrices from './unit-prices.js'
```
steve-dickinson, line 1: Just a matter of preference. Function re-written to:

```
import unitPrices from './unit-prices.js';

const calculateTierCosts = (tierInputAmounts) => {
  const tierCosts = Object.entries(tierInputAmounts).map(([tier, inputAmount]) => {
    const unitAmount = Number(inputAmount) || 0;
    const cost = unitAmount * unitPrices[tier];

    return { tier, unitAmount, cost };
  });

  const total = tierCosts.reduce((acc, cur) => acc + cur.cost, 0);

  return { tierCosts, total };
};

export default calculateTierCosts;
```

packages/webapp/src/credits/constants.js
```diff
@@ -0,0 +1,12 @@
+const routes = {
+  ESTIMATOR_CREDITS_COST: '/credits-estimation/credits-cost',
+  ESTIMATOR_CREDITS_TIER: '/credits-estimation/credits-tier'
+}
+
+export default {
+  routes,
+  views: Object.fromEntries(Object.entries(routes).map(([k, v]) => [k, v.substring(1)])),
```
steve-dickinson, line 8: Again matter of preference:

```
const views = Object.fromEntries(Object.entries(routes).map(([key, value]) => [key, value.substring(1)]));
```

steve-dickinson (commented):
Just a little fedback, but they are really just a matter of preference to make things a bit more readable. Do with it what you like :)

steve-dickinson (approved):
LGTM

shaundefra (approved):
pulled latest and ran regression test locally with no regressions

PR #422, opened 2023-09-03
feature/BNGP-3476-back--button-legal-agreement-parties

soorajvdefra:


packages/webapp/client/js/pages/land/add-legal-agreement-parties.js
```diff
@@ -37,3 +37,20 @@ $(document).ready(function (event) {
...
+  if (navigationType === 'back_forward') {
+    // eslint-disable-next-line no-undef
+    requestAnimationFrame(() => {
+      const formElement = document.querySelector('form')
+      if (formElement) {
+        formElement.reset()
+      }
+    })
+  }
+})
```
KiranVarma02, line 56: well done for making this work :). It's great that it works and we have a solution but from UI point of view only this page will behave differently when compared to other pages, I don't think its consistent, so I am not sure if bug is correct. I have added similar comment to ticket and discussed with @teddmason and @paulsimonandrews . If others agree then its ok I guess!

paulsimonandrews: I don't think this is the right approach as we should minimise as much as possible any client-side javascript.

If we pass an explicit back link to the `legal-party-list` template to point back to `add-legal-agreement-parties` this behaviour goes away. So like this 

```
return h.view(constants.views.LEGAL_PARTY_LIST, {
      backLink: constants.routes.ADD_LEGAL_AGREEMENT_PARTIES,
      legalAgreementPartiesWithAction,
      legalAgreementParties,
      legalAgreementType,
      routes: { ADD_LEGAL_AGREEMENT_PARTIES, LEGAL_PARTY_REMOVE }
    })
```

(here in the code https://github.com/DEFRA/biodiversity-net-gain-service/blob/65418257ed5719a3cf0086a2dd7f96b1908d045c/packages/webapp/src/routes/land/legal-party-list.js#L38)

However, it doesn't solve the issue of the back button. But that is one of the purposes of the back link, so the user doesn't need the back button that is difficult to effectively control the behaviour of.

shaundefra: Tech assemble we decided to close this ticket as its been superseded and the wider issue can be picked up with the rest of the team on the spike ticket [[BNGP-847] Investigate back button navigation expected behavior - JIRA (atlassian.net)](https://eaflood.atlassian.net/browse/BNGP-847)

shaundefra:
superceded with new legal agreement flow changes

PR #449, opened 2023-10-02
Feature/bngp 3637 windows defender

teddmason:
- Signalr removed
- Majority of postprocess helpers removed, bar geospatial and metric
- AV scanning managed by Microsoft defender for storage
- Files are uploaded to blob storage, service will then poll for blob tags that will get the result of the file scanning, and react to that
- File types requiring post processing (metric extraction, geospatial conversions) are then forwarded to function app
- Results returned to front end

The file scanning takes roughly 5 to 8 seconds. Timings for polling are a 4 second wait followed by a default of 20 attempts at 1s intervals for blob tags

If the service is using azurite local storage then instead of awaiting blob tags (that won't appear) the service returns mocked results for continuation.


paulsimonandrews (approved):
LGTM

paulsimonandrews (dismissed):
LGTM

paulsimonandrews (approved):
LGTM

shaundefra:
tested locally

PR #458, opened 2023-10-11
BNGP-3690 and BNGP-3712 added condition for tasklist

AjinkyaJawalkar:


usman-j-shaikh (dismissed):
LGTM

teddmason (changes requested):
This needs a unit test to cover the logic for the startTaskUrl in task list object.

teddmason (approved):

PR #472, opened 2023-10-23
BNGP-3780 dependency update sweep

teddmason:


teddmason:
Dependencies not updating:

- gdal-async; latest still has maintainer comments on it without closing a string correctly
- signal-r; latest breaks azure deployments (removed with defender work anyway)
- lerna; latest version breaks our package installations

pwadmore-ea (approved):

PR #497, opened 2023-11-06
boilerplate

teddmason:


packages/webapp/src/routes/land/clients-email-address.js
```diff
@@ -1,13 +1,36 @@
...
 const handlers = {
   get: async (request, h) => {
-    return h.view(constants.views.CLIENTS_EMAIL_ADDRESS)
+    const email = request.yar.get(constants.redisKeys.CLIENTS_EMAIL_ADDRESS)
+    return h.view(constants.views.CLIENTS_EMAIL_ADDRESS, {
+      email
+    })
+  },
+  post: async (request, h) => {
+    const email = request.payload.email
```
KiranVarma02, line 12: I prefer destructing but its up to you :) 
```suggestion
    const { email } = request.payload
```

packages/webapp/src/routes/land/clients-email-address.js
```diff
@@ -1,13 +1,36 @@
...
+    const error = validateEmail(email, '#email')
+    if (error) {
+      if (error.err[0].text === 'Enter your email address') {
+        error.err[0].text = 'Enter email address'
```
KiranVarma02, line 16: I don't have access to CD. Does this need to be different?

teddmason: Yes it's a bit of a fudge, but thought it would be better than over-complicating it, but it's because there is an email validation helper already used for several other routes that ask for the user's email, where as the copy for this ticket is not for the user's email.

packages/webapp/src/routes/land/clients-email-address.js
teddmason: Think I have the habit of destructuring if there are multiple values being pulled, but not if there is only 1, happy to update

KiranVarma02 (approved):

PR #507, opened 2023-11-13
feature/BNGP-3826-bug-duplicate-entries-legal-agreement

soorajvdefra:


packages/webapp/src/routes/land/add-planning-authority.js
```diff
@@ -39,7 +42,21 @@ const handlers = {
...
+      '#localPlanningAuthority',
+      'This local planning authority has already been added - enter a different local planning authority, if there is one',
+      excludeIndex
+    )
+    if (duplicateError) {
+      return h.view(constants.views.ADD_PLANNING_AUTHORITY, {
+        err: Object.values(duplicateError),
+        localPlanningAuthorityNameErr
+      })
+    }
```
KiranVarma02, line 59: if lpaList.includes(selectedLpa) then its duplicate right? I guess this not cover all scenarios?

packages/webapp/src/utils/helpers.js
```diff
@@ -419,6 +419,42 @@ const validateTextInput = (text, hrefId, fieldType = 'input', maxLength = null,
...
+  })
+  if (duplicate) {
+    error.err = {
+      text: errorMessage,
+      href: hrefId
+    }
+    return error
+  }
+  return null
+}
```
KiranVarma02, line 457: If we are going with this approach ( I am not sure yet) then we need tests here

KiranVarma02 (approved):
LGTM but please get other opinions

packages/webapp/src/utils/helpers.js
pwadmore-ea, line 457: @soorajvdefra  - I agree with @KiranVarma02 on the need for tests here. Additionally, I think these functions need commenting. I've managed to reach the generic error page when adding landowners for conservation covenants by performing the following actions:

- Add an organisation
- Add an individual
- Add another organisation

Attempted addition of the second organisation causes  **"statusCode":500,"message":"item.toLowerCase is not a function"**.

pwadmore-ea (changes requested):

packages/webapp/src/routes/land/add-planning-authority.js
soorajvdefra: as per discussion excludeIndex required for edit functionality

packages/webapp/src/utils/helpers.js
soorajvdefra: yes added the tests. didnt know there was a test file

soorajvdefra: fixed defect, added tests and comment

pwadmore-ea (dismissed):

LeenaKu:
Tested and Passed.

pwadmore-ea (approved):

shaundefra:
tested by leena

PR #530, opened 2023-12-03
fix/BNGP-4099 show metric error for non-metric spreadsheet

paulsimonandrews:
https://eaflood.atlassian.net/browse/BNGP-4099


packages/webapp/src/routes/__tests__/land/upload-metric.spec.js
```diff
@@ -173,12 +170,33 @@ describe('Metric file upload controller tests', () => {
       })
     })
 
+    it('should return error if valid spreadsheet is not a valid metric', (done) => {
```
KiranVarma02, line 173:
```suggestion
    it('should return error if spreadsheet is not a valid metric', (done) => {
```

KiranVarma02 (approved):

shaundefra:
tested locally

PR #551, opened 2023-12-19
feature/BNGP-3696-malicious-file-processing

pwadmore-ea:
https://eaflood.atlassian.net/browse/BNGP-3696


packages/azure-functions/ProcessMaliciousUploads/index.mjs
```diff
@@ -0,0 +1,89 @@
...
+let itemProcessingTimeout
+let processingTimeout
+
+export default async function (context, _timer) {
+  context.log('Processing malicious uploads')
+  let pauseProcessing = false
+  // If all malicious ulploads are not processed within a configurable period of time,
+  // pause processing unti the next scheduled invocation of this function.
+  const timeoutId = setTimeout(() => { pauseProcessing = true }, getProcessingTimeout(context))
+  for await (const maliciousUpload of await blobStorageConnector.findBlobsInContainerByTags(config)) {
```
KiranVarma02, line 22: could we extract `await blobStorageConnector.findBlobsInContainerByTags(config)` to const just to simplify line 22?

packages/azure-functions/ProcessMaliciousUploads/index.mjs
```diff
@@ -0,0 +1,89 @@
...
+
+const containerName = 'customer-uploads'
+const tags = `"${MALWARE_SCANNNING_RESULT_TAG_NAME}" = '${MALWARE_SCANNNING_RESULT_TAG_VALUE}'`
+const config = {
+  containerName,
+  tags
+}
+
+let itemProcessingTimeout
+let processingTimeout
```
KiranVarma02, line 14: does this need to be at this scope? its only used in 1 function

pwadmore-ea: These variables are at module level to facilitate lazy initialisation. The associated environment variables are parsed on first use and the resulting timeouts are stored in these variables.

packages/azure-functions/ProcessMaliciousUploads/index.mjs
pwadmore-ea: Yes but it is not vital.

KiranVarma02 (dismissed):

PR #559, opened 2023-12-28
 feature/BNGP-4219-suppress-metric-validation

soorajvdefra:


packages/webapp/src/utils/helpers.js
```diff
@@ -613,7 +613,10 @@ const getMetricFileValidationErrors = (metricValidation, href, useStatutoryMetri
   } else if (!metricValidation.isOffsiteDataPresent) {
     error.err[0].text = 'The selected file does not have enough data'
   } else if (!metricValidation.areOffsiteTotalsCorrect) {
-    error.err[0].text = 'The selected file has an error - the baseline total area does not match the created and enhanced total area for the off-site'
+    console.log('sooraj')
```
paulsimonandrews, line 616: Remove the console log

paulsimonandrews (commented):
Tiny change to remove console log.

Code looks fine, before approving/merging would like to check with Chris that the validation will be re-introduced. If definitely not, we should remove it properly, not just disable as the code does here.

packages/webapp/src/utils/helpers.js
soorajvdefra: console.log removed

paulsimonandrews (approved):
LGTM

shaundefra:
tested locally

PR #561, opened 2023-12-29
feature/BNGP-4130-relax-metric-check-and-submit-validation

soorajvdefra:


paulsimonandrews (commented):
Code is fine and implements what is on the current ticket.

Before approval/merging, want to check the behaviour if a value isn't there (should we insert a blank value, or not add the field to the payload as happens here)

KiranVarma02 (approved):

PR #593, opened 2024-01-18
fix: BNGP-4370 restrict exposedHeaders

KiranVarma02:


packages/webapp/src/server.js
```diff
@@ -24,7 +25,6 @@ const createServer = async options => {
             abortEarly: false
           }
         },
-        cors: true,
         security: true
```
pwadmore-ea, line 28: I think security:true can be removed as well because the custom response headers replace what cors:true and security:true provide.

packages/webapp/src/server.js
```diff
@@ -47,6 +47,24 @@ const init = async server => {
...
+    options: {
+      keys: [
+        { key: 'X-Frame-Options', value: 'deny' },
+        { key: 'X-Content-Type-Options', value: 'nosniff' },
+        { key: 'Access-Control-Allow-Origin', value: SERVICE_HOME_URL },
+        { key: 'Cross-Origin-Opener-Policy', value: 'same-origin' },
+        { key: 'Cross-Origin-Embedder-Policy', value: 'require-corp' },
+        { key: 'X-Robots-Tag', value: 'noindex, nofollow' },
+        { key: 'X-XSS-Protection', value: '1; mode=block' },
+        { key: 'Strict-Transport-Security', value: 'max-age=31536000;' },
```
pwadmore-ea, line 61: Please use the default max-age configured by Hapi.js - https://hapi.dev/api/?v=21.3.2#-routeoptionssecurity

pwadmore-ea (changes requested):

packages/webapp/src/server.js
KiranVarma02: ok, done

packages/webapp/src/server.js
KiranVarma02: done

pwadmore-ea (dismissed):

pwadmore-ea (dismissed):

pwadmore-ea (approved):

shaundefra:
tested locally

PR #619, opened 2024-02-07
SEMVER PATCH fix/BNGP-4555-habitat-reference

pwadmore-ea:
https://eaflood.atlassian.net/browse/BNGP-4555

Handle non-string habitat references


packages/webapp/src/utils/application.js
```diff
@@ -187,7 +187,7 @@ const getHabitats = session => {
 
   const proposed = proposedIdentifiers.flatMap(identifier =>
     metricData[identifier].filter(details => 'Condition' in details).map(details => ({
-      proposedHabitatId: details['Habitat reference Number'],
+      proposedHabitatId: details['Habitat reference Number'] ? String(details['Habitat reference Number']) : details['Habitat reference Number'],
```
KiranVarma02, line 190:
```suggestion
      proposedHabitatId: details['Habitat reference Number'] ?? String(details['Habitat reference Number']),
```

does this work?

paulsimonandrews: Suggested change won't convert from number to string unfortunately, but the original will

paulsimonandrews (approved):
I think this is the best approach for the fix. I had considered returning an empty string if null/undefined, but that will change the behaviour more than the current proposed solution that will behave the same as has been tested, but just convert Number to String if needed.

shaundefra:
tested locally

PR #638, opened 2024-02-18
BNGP-4128 & BNGP-4606: Credits purchase task list and session saved to database

paulsimonandrews:
https://eaflood.atlassian.net/browse/BNGP-4128


packages/database-version-control/src/migrations/down/2024.03.07T00.00.00.alter-credits-app-reference-function.sql
```diff
@@ -0,0 +1,48 @@
+DROP FUNCTION bng.fn_create_credits_app_reference (UUID, bng.APPLICATION_TYPE, UUID, INTEGER);
+CREATE FUNCTION bng.fn_create_credits_app_reference
```
KiranVarma02, line 2: This is first time we are creating this DB function right?

I thought down sql will only have drop function (not create) to get DB back to its original state?

KiranVarma02 (approved):

PR #640, opened 2024-02-19
feat: BNGP-4140 Remove eligibility routes

KiranVarma02:


packages/webapp/src/routes/__tests__/test/seed-developer-data.spec.js
paulsimonandrews, line 1: I don't think we need to remove this test, just update to not check the two values that have been removed. We'll still want the seed data for the developer journey, so the test is also useful.

We'll probably want to update the actual seed data to remove the data associated with the redis key removal in this PR

paulsimonandrews (changes requested):
Couple of things to check/look-at

1. I think you've missed removing some of the routes javascript files:

developer/details-name
developer/eligibility-*

2. I don't think you should remove the seed data test, but update the seed data and test as appropriate to remove the removed redis keys/data.

3. Can you search the code base for the redis keys that you are removing to see if there are any other dependencies? A quick check shows there is a function in `helpers.js` that could be removed: `getDeveloperEligibilityResults()`

packages/webapp/src/routes/__tests__/test/seed-developer-data.spec.js
KiranVarma02: reverted

packages/webapp/src/routes/developer/routing-result.js
```diff
@@ -2,7 +2,7 @@ import constants from '../../utils/constants.js'
 
 const handlers = {
   get: async (_request, h) => h.view(constants.views.DEVELOPER_ROUTING_RESULT),
-  post: async (_request, h) => h.redirect(constants.routes.DEVELOPER_ELIGIBILITY_ENGLAND)
+  post: async (_request, h) => h.redirect(constants.routes.DEVELOPER_ROUTING_RESULT)
```
KiranVarma02, line 5: @paulsimonandrews - do you know what this route is? I just put it to redirect to itself but where should it redirect to?

paulsimonandrews: I think this is fine for the moment. These routes look like they aren't needed any more anyway so will need to removed, but that isn't in scope here.

paulsimonandrews (approved):

paulsimonandrews (approved):
LGTM

shaundefra:
tested locally

PR #646, opened 2024-02-23
Feat/bngp 4581 credits estimator header link

jd-barnard:


packages/webapp/src/views/layout.html
```diff
@@ -55,7 +55,7 @@
     homepageUrl: "/",
     containerClasses: "govuk-width-container",
     serviceName: serviceName,
-    serviceUrl: "/"
+    serviceUrl: serviceUrl
```
paulsimonandrews, line 58: `serviceUrl: serviceUrl or "/"` should give us a default that can be overriden by a `serviceUrl` if one is set

paulsimonandrews (changes requested):
Small improvement suggested

packages/webapp/src/views/layout.html
jd-barnard: updated

paulsimonandrews (approved):

shaundefra:
tested locally

PR #649, opened 2024-02-26
BNGP-3621: Non relevant persons: upload letter of consent and authority

jd-barnard:


packages/webapp/src/routes/__tests__/developer/consent-agreement-check.spec.js
```diff
@@ -1,7 +1,7 @@
 import constants from '../../../utils/constants.js'
 import { submitPostRequest } from '../helpers/server.js'
 
-const url = '/developer/check-consent-file'
+const url = '/developer/check-consent-file-test'
```
KiranVarma02, line 4: you added test to differentiate from main route? I think I prefer to use same constant as main route to not have to maintain path in 2 places.

KiranVarma02: I see you used main constant in other test files, so I guess this is a mistake!

jd-barnard: this test already existed with the route - but I've only added that route in now with this PR. There's a test in this file which expects a 404. On reflection that sounds like a bit of a side effect (the route didn't exist when that test was written). So I should probably just remove that test?

KiranVarma02: ahh ok, I think I would remove test

KiranVarma02 (approved):

KiranVarma02 (approved):

shaundefra:
tested by Tajinder

PR #656, opened 2024-03-01
BNGP-4652 Make footer "Terms and Conditions" link journey specific

jd-barnard:
…e credits terms and conditions

### DIFF
--- a/packages/webapp/src/views/footer.html
+++ b/packages/webapp/src/views/footer.html
@@ -8,7 +8,7 @@
           href: "/cookies",
           text: "Cookies"
         },{
-          href: "https://www.gov.uk/guidance/biodiversity-gain-sites-register-terms-and-conditions",
+          href: termsAndConditions or "https://www.gov.uk/guidance/biodiversity-gain-sites-register-terms-and-conditions",
           text: "Terms and conditions"
         },{
           href: "/accessibility-statement",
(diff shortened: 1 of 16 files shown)


packages/webapp/src/views/footer.html
```diff
@@ -8,7 +8,7 @@
           href: "/cookies",
           text: "Cookies"
         },{
-          href: "https://www.gov.uk/guidance/biodiversity-gain-sites-register-terms-and-conditions",
+          href: termsAndConditions or "https://www.gov.uk/guidance/biodiversity-gain-sites-register-terms-and-conditions",
```
KiranVarma02, line 11: does this look right? or a mistake?

jd-barnard: This is using the termsAndConditions variable - if it's not set then it will default to the string. I've set that variable on all the credits views to use the credits terms and conditions. All other pages are falling back to the default string. So I can't see a mistake.

KiranVarma02 (approved):

KiranVarma02 (approved):

KiranVarma02 (approved):

paulsimonandrews (approved):
LGTM

shaundefra:
tested by Tajinder

PR #657, opened 2024-03-01
BNGP-4650 Validation: Credits Estimator apply credits purchase validation to the credits estimator tool

jd-barnard:


packages/webapp/src/routes/credits-estimation/credits-tier.js
```diff
@@ -1,9 +1,41 @@
...
+const handlers = {
+  get: async (request, h) => {
+    const previousCostCalculation = request.yar.get(creditsEstimationConstants.redisKeys.ESTIMATOR_CREDITS_CALCULATION)
+    const inputValues = (previousCostCalculation)
+      ? Object.fromEntries(previousCostCalculation.tierCosts.map(({ tier, unitAmount, _ }) => [tier, unitAmount]))
+      : {}
+    return h.view(creditsEstimationConstants.views.ESTIMATOR_CREDITS_TIER, { inputValues })
+  },
+  post: async (request, h) => {
+    request.yar.set(creditsEstimationConstants.redisKeys.ESTIMATOR_CREDITS_CALCULATION, calculateCost(request.payload))
```
KiranVarma02, line 19: just for my understanding, do you know if value is set before validation or after validation? after validation i guess!

jd-barnard: the value is set only if validation passes - if validation fails then an Error is thrown (see helpers line 801).

KiranVarma02 (approved):

shaundefra:
tested locally

PR #661, opened 2024-03-04
BNGP-4627: Date entry field bug

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-4627

This change fixes a bug on mutiple pages where non-numeric characters on date entry fields were not being handled with an appropriate error message.It also removes the unwanted behaviour of adding a zero when an invalid entry has been made by the user.


packages/webapp/src/routes/__tests__/credits-purchase/date-of-birth.spec.js
```diff
@@ -76,7 +82,7 @@ describe(url, () => {
       postOptions.payload['dob-year'] = goodYear
       const res = await submitPostRequest(postOptions, 200)
       expect(res.payload).toContain('There is a problem')
-      expect(res.payload).toContain('Date of birth must be a real date')
+      expect(res.payload).toContain('Date of birth must be a number')
```
soorajvdefra, line 85: Date of birth must be a number is not the correct error message. old message is the correct one

packages/webapp/src/routes/__tests__/credits-purchase/date-of-birth.spec.js
```diff
@@ -85,7 +91,7 @@ describe(url, () => {
...
```
soorajvdefra, line 94: Date of birth must be a number is not the correct error message. old message is the correct one

thousand-leaves:
closing because of duplication by https://github.com/DEFRA/biodiversity-net-gain-service/pull/664

PR #666, opened 2024-03-06
BNGP-2773: Fix js console error on add-legal-agreement-parties page

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-2773

This changes fixes the javascript console error on the add-legal-agreement-parties page


packages/webapp/src/views/land/add-legal-agreement-parties.html
```diff
@@ -107,5 +107,4 @@ <h1 class="govuk-heading-l" tabindex="-1">
 {% endblock %}
 {% block bodyEnd %}
     {{ super() }}
-    <script src="{{ assetPath }}/js/addLegalAgreementParties.js"></script>
 {% endblock %}
```
KiranVarma02, line 110: There are couple of other places in the codebase we follow same pattern (email-entry.html, check-and-submit.html, check-land-boundary-details.js), just want to understand if console errors are only on this page?

if this is the fix we are going with, do we still need lines 108-110? Could we remove all block?

thousand-leaves: the error was caused by a deprecated bit of code in the public/build/addLegalAgreementParties.js file which is no longer needed (so I've removed) and doesn't appear any where else in the code, apart from in the land/add-legal-agreement-parties.js but this is not causing a console error.

yes i think we could remove the whole block.

KiranVarma02 (approved):

PR #668, opened 2024-03-11
BNGP-3494 Move application reference generation out of Postgres function into Javascript/Node

jd-barnard:


packages/azure-functions/Shared/reference-helpers.js
```diff
@@ -0,0 +1,28 @@
...
+  const chars = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+  let result = ''
+  for (let i = 0; i < length; i++) {
+    const randomIndex = Math.floor(randomBytes(1)[0] / 255 * chars.length)
+    result += chars[randomIndex]
+  }
+  return result
+}
+
+const retry = async (query, options, retries = 5) => {
```
KiranVarma02, line 13: (I guess you named retry same like other existing db operation query functions)

I am not sure but would `retryDbOperation` read better?!!

jd-barnard: That's a good suggestion - 'retry' didn't seem quite right to me either (but I couldn't think of anything else!)

jd-barnard: @KiranVarma02  function name updated

KiranVarma02 (approved):

KiranVarma02 (dismissed):

PR #671, opened 2024-03-13
BNGP-4691: Remove redundant pages from landowner journey

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-4691

This change removes redundant pages, views, routes and constants from the registration/landowner journey.


jd-barnard (approved):

shaundefra:
tested by Leena

PR #676, opened 2024-03-15
feature/BNGP-4125-cya-pages-null-handling

soorajvdefra:
following issue also fixed by this PR
https://eaflood.atlassian.net/browse/BNGP-4522


packages/webapp/src/routes/land/check-legal-agreement-details.js
```diff
@@ -20,6 +21,11 @@ const handlers = {
     }, {
       inProgressUrl: constants.routes.CHECK_LEGAL_AGREEMENT_DETAILS
     })
+
+    const registrationTaskStatus = getIndividualTaskStatus(request.yar, REGISTRATIONCONSTANTS.LEGAL_AGREEMENT)
+    if (registrationTaskStatus !== 'COMPLETED') {
+      return h.redirect(constants.routes.REGISTER_LAND_TASK_LIST)
+    }
```
KiranVarma02, line 28: looks like we do this check only in this case to check status, why is that?

KiranVarma02 (approved):

packages/webapp/src/routes/land/check-legal-agreement-details.js
soorajvdefra: discussed

shaundefra:
tested by Tajinder

PR #679, opened 2024-03-18
feature/BNGP-4739-credit-purchase-remove-unncessary-update

soorajvdefra:


packages/webapp/src/views/credits-purchase/check-and-submit.html
```diff
@@ -21,6 +21,7 @@
     'valueId': 'metric-development-details-confirmed-value',
     'href': metric.detailsConfirmedUrl,
     'visuallyHiddenText': ' confirm development details',
+    'actionClasses':'hidden',
```
paulsimonandrews, line 24: You should use the newly added`'noAction': true` option that stops the link being added within the common template rather than just hiding with CSS

paulsimonandrews (commented):
Use `'noAction': true` (see inline comment)

KiranVarma02 (approved):

shaundefra:
tested locally

PR #681, opened 2024-03-20
BNGP-4586 Update task list to reflect V23 of prototype

jd-barnard:


thousand-leaves (approved):

shaundefra:
tested locally : FAIL in its current form but merging to unblock dependent tickets that will resolve the tasklist sections that are missing fyi @paulsimonandrews

PR #682, opened 2024-03-21
Update credits purchase queue name to match infra in TST and above

paulsimonandrews:


shaundefra (approved):
have seen the changes with @paulsimonandrews

PR #688, opened 2024-03-25
BNGP-4771: Credits payload metric Number-String bug

paulsimonandrews:


packages/webapp/src/routes/credits-purchase/check-and-submit.js
```diff
@@ -58,6 +58,8 @@ const handlers = {
     const claims = request.auth.credentials.account.idTokenClaims
     const { currentOrganisation } = getOrganisationDetails(claims)
 
+    // test comment
```
jd-barnard, line 61: should this be removed?

jd-barnard (approved):

jd-barnard (approved):

packages/webapp/src/routes/credits-purchase/check-and-submit.js
paulsimonandrews: Yes and done

PR #692, opened 2024-03-27
BNGP-4714: Fix bug on Check Defra Account paragraph 2 not suppressed

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-4714

This change fixes a bug found on the check DEFRA account details page on the land journey where the second paragraph should be suppressed if user is an agent but the paragraph is not being suppressed.


KiranVarma02 (approved):

shaundefra:
tested by Tajinder

PR #696, opened 2024-04-03
BNGP-4781 Investigate replacing Redis with in-memory caching for sessions

jd-barnard:
…rences to redis to be clearer we are using cache


docker/azure-services/local-webapp-env-template
```diff
@@ -1,8 +1,6 @@
 # WARNING - All environment variables MUST be enclosed in double quotes to ensure the maximum degree of protection
 # against secrets being committed to version control unintentionally.
 export SERVER_PORT="3000"
-export REDIS_HOST="redis"
-export REDIS_PORT="6379"
 export SESSION_COOKIE_PASSWORD="the-password-must-be-at-least-32-characters-long"
```
KiranVarma02, line 4: Could you please add below in this file?
```
CACHE_MAX_BYTE_SIZE, CACHE_CLEANUP_INT_MIN_SEC, CACHE_CLONE_BUFFER_ON
```

packages/webapp/src/plugins/cache.js
```diff
@@ -1,17 +1,22 @@
...
+const options = {}
+if (CACHE_MAX_BYTE_SIZE) {
+  options.maxByteSize = CACHE_MAX_BYTE_SIZE
+}
+if (CACHE_CLEANUP_INT_MIN_SEC) {
+  options.minCleanupIntervalMsec = CACHE_CLEANUP_INT_MIN_SEC
+}
+if (CACHE_CLONE_BUFFER_ON) {
+  options.cloneBuffersOnGet = CACHE_CLONE_BUFFER_ON
+}
```
KiranVarma02, line 13: I am interested - could you please send me a docs link on what these values should be in local, prod?

jd-barnard: Sure (they were already in the jira as well): https://hapi.dev/module/catbox-memory/api/?v=6.0.1#new-catboxmemoryengineoptions.

You don't need to set these if you are happy with the defaults.

docker/azure-services/local-webapp-env-template
jd-barnard: These are not required as it will fall back to defaults.

PR #720, opened 2024-04-29
BNGP 4518 metric to include watercourses (re-apply after revert)

paulsimonandrews:


jd-barnard (approved):

shaundefra:
tested locally, no visible regressions

PR #721, opened 2024-04-29
feature/BNGP-4896-creditpurchase-focus-control

soorajvdefra:


KiranVarma02 (commented):
Would normal approach of focus which is done in other places not work here?

packages/webapp/src/views/credits-purchase/check-and-submit.html
```diff
@@ -57,7 +57,7 @@
     'value': credits.amounts.a1,
     'valueDataTestId': 'credits-a1-value',
     'valueId': 'credits-a1-value',
-    'href': credits.changeUrl,
+    'href': credits.changeUrl + '?focus=a1-units',
```
KiranVarma02, line 60: I have just tested and this works perfect without any js. Please try and let me know

` 'href': credits.changeUrl+'#a1-units',`

packages/webapp/src/views/credits-purchase/check-and-submit.html
```diff
@@ -66,7 +66,7 @@
     'value': credits.amounts.a2,
     'valueDataTestId': 'credits-a2-value',
     'valueId': 'credits-a2-value',
-    'href': credits.changeUrl,
+    'href': credits.changeUrl + '?focus=a2-units',
```
KiranVarma02, line 69: `'href': credits.changeUrl+'#a2-units',`

KiranVarma02 (changes requested):

packages/webapp/src/views/credits-purchase/check-and-submit.html
soorajvdefra: Yes, Kiran, your suggestion should work. However, there's another issue: using 'href': credits.changeUrl+'#a1-units' will focus on the element with scrolling. If the user needs to confirm which box it is, they will need to scroll up!. I will demonstrate this tomorrow.

KiranVarma02 (approved):

KiranVarma02 (approved):

shaundefra:
tested locally

PR #744, opened 2024-05-21
Fix/BNGP 5031 update npm packages to address latest vulnerabilities

Stewart-Jumbe:
Update to package.json files to address the latest vulnerabilities

https://eaflood.atlassian.net/browse/BNGP-5031

Dependabot identified vulnerabilities in the code, this branch addresses them


packages/azure-functions/Shared/reference-helpers.js
```diff
@@ -4,7 +4,7 @@ const randomReferenceString = (length) => {
   const chars = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
   let result = ''
   for (let i = 0; i < length; i++) {
-    const randomIndex = Math.floor(randomBytes(1)[0] / 255 * chars.length)
+    const randomIndex = Math.floor(randomBytes(1)[0] / 256 * chars.length)
```
jd-barnard, line 7: Good spot Stewart! Paul suggested instead using `crypto.randomInt(0, chars.length)`. I agree that is a simpler and better approach - it should give us a more random result.

Stewart-Jumbe: Thanks John, will do

packages/azure-functions/package.json
```diff
@@ -29,7 +29,9 @@
     "@defra/bng-geoprocessing-service": "1.85.0",
     "@defra/bng-metric-service": "1.85.0",
     "@defra/bng-utils-lib": "1.85.0",
-    "moment": "^2.29.4"
+    "moment": "^2.29.4",
+    "ip": ">=2.0.1"
```
jd-barnard, line 33: why is this being added?

packages/geoprocessing-service/package.json
```diff
@@ -21,8 +21,10 @@
   "license": "OGL-UK-3.0",
   "dependencies": {
     "@defra/bng-errors-lib": "1.85.0",
-    "gdal-async": "^3.8.3",
-    "geodesy": "^2.4.0"
+    "gdal-async": "^3.8.5",
+    "geodesy": "^2.4.0",
+    "ip": ">=2.0.1"
```
jd-barnard, line 26: why is this being added?

packages/azure-functions/package.json
jd-barnard: I think ip is a dependency of lerna and so instead perhaps it is lerna that needs upgrading?

jd-barnard (approved):

PR #751, opened 2024-05-24
Fix package-lock to include typescript so versioning works again

paulsimonandrews:


shaundefra (approved):

shaundefra:
approved with @paulsimonandrews

PR #768, opened 2024-06-03
BNGP-5106: Fix terms and conditions acceptance not validated

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5106

This change applies validation to the terms and conditions check box so that a user can't submit an application unless they have marked the checkbox as read.


jd-barnard (approved):

shaundefra:
tested locally

PR #770, opened 2024-06-06
BNGP-5108 Regression - User should not be redirected to the tasklist if going back (in the browser) after submission

jd-barnard:
…to manage biodiversity gains


thousand-leaves (dismissed):

thousand-leaves (approved):

shaundefra:
tested locally

PR #771, opened 2024-06-06
Feat/bngp 5126 post redirect get

jd-barnard:


packages/webapp/src/plugins/redirect-view.js
```diff
@@ -0,0 +1,37 @@
+import constants from '../utils/constants.js'
+
+const redirectView = {
+  name: 'redirect-view',
+  register: (server, _options) => {
+    server.decorate('toolkit', 'redirectView', function (route, data) {
+      if (this.request.isInjected && route !== '/test') {
```
jd-barnard, line 7: isInjected is where we inject via a test - the exception here is the /test route which is used by the redirect-view.spec.js test to assert the redirection happens. i.e. we don't want a redirect for the existing error view tests on existing routes.

packages/webapp/src/utils/redirect-view-handler.js
```diff
@@ -0,0 +1,8 @@
+const addRedirectViewUsed = (handler) => {
```
jd-barnard, line 1: this is preferred to having to add the boolean to all routes that return an error following a form submission - i.e. it would be easy to miss.

thousand-leaves (dismissed):

packages/webapp/src/utils/constants.js
```diff
@@ -186,6 +187,7 @@ const redisKeys = {
   SAVE_APPLICATION_SESSION_ON_SIGNOUT_OR_JOURNEY_CHANGE,
   PRE_AUTHENTICATION_ROUTE,
   SAVE_APPLICATION_SESSION_ON_SIGNOUT,
+  VIEW_DATA
```
StuAA78, line 190: Missing a comma.
```suggestion
  VIEW_DATA,
```

packages/webapp/src/routes/land/add-hectares.js
```diff
@@ -1,4 +1,6 @@
 import constants from '../../utils/constants.js'
+import { getValidReferrerUrl } from '../../utils/helpers.js'
```
StuAA78, line 2: I couldn't see `getValidReferrerUrl` being used in this file?

packages/webapp/src/routes/land/clients-name.js
```diff
@@ -1,5 +1,6 @@
 import constants from '../../utils/constants.js'
-import { validateFirstLastNameOfLandownerOrLeaseholder } from '../../utils/helpers.js'
+import { getValidReferrerUrl, validateFirstLastNameOfLandownerOrLeaseholder } from '../../utils/helpers.js'
```
StuAA78, line 2: I couldn't see `getValidReferrerUrl` being used in this file?

packages/webapp/src/routes/land/uk-address.js
```diff
@@ -1,5 +1,6 @@
 import constants from '../../utils/constants.js'
-import { validateAddress } from '../../utils/helpers.js'
+import { redirectAddress, validateAddress } from '../../utils/helpers.js'
```
StuAA78, line 2: I couldn't see `redirectAddress` being used in this file?

StuAA78 (changes requested):
Spotted a missing comma in the array of redis keys in `constants.js` that may cause issues? Also a few cases where we're importing a file but don't seem to use it.

Other than that all looks good to me 😃

StuAA78 (dismissed):
👍

thousand-leaves (approved):

PR #780, opened 2024-06-13
BNGP-5149: Add enable combined case env var

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5149

This change adds the enable combined case environment variable as a feature flag and updates relevant files required for setup of the combined case journey.


paulsimonandrews (approved):
LGTM

PR #782, opened 2024-06-14
BNGP-5160: Create dashboard and application list page for combined case

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5160

This change adds an interim link to the main dashboard for combined case in order to assist with development. It also adds an application list page so that a journey can be started.


paulsimonandrews (dismissed):
LGTM

paulsimonandrews (approved):
LGTM

PR #791, opened 2024-06-19
BNGP-5191: Start page for register or combined case

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5191

This change add a start page for the register or combined case journey.


paulsimonandrews (approved):
LGTM

PR #793, opened 2024-06-25
BNGP-5210: fix referer logic for changing metric upload

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5210

Fixes https://eaflood.atlassian.net/browse/BNGP-5201

The current referer logic means the upload metric file pages are skipped when going back to change the metric file from the check metric details page in the developer journey. We fix this by updating the logic to no longer skip the pages if coming from the `developer/confirm-off-site-gain` route.


paulsimonandrews (approved):
LGTM

PR #794, opened 2024-06-26
BNGP-5184: Remove query code from allocation API call

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5184

APIM will inject query code used in API call if it isn’t present, so we only want to inject it ourselves based on the value in an environment variable. This PR does this by removing the default value for the `BACKEND_API.CODE_QUERY_PARAMETER` config item, and makes setting the query param conditional based on whether the config item has a value.


jd-barnard (approved):

shaundefra:
tested by Leena

PR #797, opened 2024-06-27
BNGP-5077: Remove 'optional' property from payloads

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5077

An 'optional' property for files is appearing in the registration and allocation journey payloads, which does not match the agreed schema. This change filters out the property from the payloads by adding some additional logic to the application file.


shaundefra:
tested locally

PR #801, opened 2024-07-05
BNGP-5233 Add interim task list (as per design)

jd-barnard:


packages/webapp/src/utils/combined-case-constants.js
```diff
@@ -1,12 +1,224 @@
...
+  COMBINED_CASE_CHECK_LANDOWNERS: reusedRoutePath(baseUrl, CHECK_LANDOWNERS),
+  COMBINED_CASE_ENHANCEMENT_WORKS_START_DATE: reusedRoutePath(baseUrl, ENHANCEMENT_WORKS_START_DATE),
+  COMBINED_CASE_UPLOAD_HABITAT_PLAN: reusedRoutePath(baseUrl, UPLOAD_HABITAT_PLAN),
+  COMBINED_CASE_CHECK_HABITAT_PLAN_FILE: reusedRoutePath(baseUrl, CHECK_HABITAT_PLAN_FILE),
+  COMBINED_CASE_HABITAT_ENHANCEMENTS_END_DATE: reusedRoutePath(baseUrl, HABITAT_ENHANCEMENTS_END_DATE),
+  COMBINED_CASE_ADD_PLANNING_AUTHORITY: reusedRoutePath(baseUrl, ADD_PLANNING_AUTHORITY),
+  COMBINED_CASE_CHECK_PLANNING_AUTHORITIES: reusedRoutePath(baseUrl, CHECK_PLANNING_AUTHORITIES)
+}
+
+export {
```
StuAA78, line 217: Just checking the reason for this being `export { ... }` when it was previously `export default { ... }`? If we reverted to `export default` then would that mean we could just use `import combinedCaseConstants` elsewhere instead of `import * as combinedCaseConstants`?

StuAA78 (commented):
Great stuff 🙂 Just a few queries before I hit approve; there are some tests marked as `.skip` and `.only` and I wanted to confirm that was intentional; plus there's a change to how we export `combined-case-constants.js` and I just wanted to check the reason for the change?

StuAA78 (dismissed):
:shipit:

StuAA78 (dismissed):
🙌

StuAA78 (approved):
👍

shaundefra:
tested by leena

PR #802, opened 2024-07-08
BNGP-5239: Create habitat matching page

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5239

This change creates the habitat matching page by adding a new route, view (and tests) based on the prototype design.


packages/webapp/src/routes/combined-case/match-habitats.js
```diff
@@ -0,0 +1,70 @@
...
+  )
+})
+
+const handlers = {
+  get: async (request, h) => {
+    // TODO: uncomment below to replace mockMetricData with session data when the rest of the journey is built
+    // const registrationMetricData = request.yar.get(constants.redisKeys.METRIC_DATA)
+    // const allocationMetricData = request.yar.get(constants.redisKeys.DEVELOPER_METRIC_DATA)
+    const { page = '1' } = request.query
+    const currentPage = parseInt(page, 10)
```
paulsimonandrews, line 38: The second param is redundant here as the radix defaults to 10 anyway

packages/webapp/src/routes/combined-case/match-habitats.js
```diff
@@ -0,0 +1,70 @@
+import constants from '../../utils/constants.js'
+import getHabitatType from '../../utils/getHabitatType.js'
+import mockMetricData from './mock-metric-data.js'
+
+const habitatKeys = ['d2', 'd3', 'e2', 'e3', 'f2', 'f3']
+
+export const getNumberOfPages = (data) => habitatKeys.reduce((acc, key) => acc + data[key].slice(0, -1).length, 0)
```
paulsimonandrews, line 7: Just a note, nothing to change.

Good for now, but this might change as not all allocation habitats will have something match against. But we need to work that out later when we do the habitat matching logic

paulsimonandrews (approved):
LGTM

Couple of comments, but nothing that can't or won't be addressed in follow-on tickets

shaundefra:
tested locally with caveats

PR #804, opened 2024-07-09
BNGP-5249 Clear metric checked session key when user checks 'no, upload again' after metric upload

jd-barnard:


thousand-leaves (approved):

shaundefra:
not able to replicate issue without fix so closing pr and deleting this branch

PR #805, opened 2024-07-09
BNGP-5221 Amendments and Removals - dev work on FE service to add links to guidance on dashboard page(s) and start page

jd-barnard:
…var for both URLs with a fallback

### DIFF
--- a/packages/webapp/src/views/common/applications.html
+++ b/packages/webapp/src/views/common/applications.html
@@ -54,7 +54,7 @@
             </tbody>
           </table>
         {%if showAmendmentGuidanceLink %}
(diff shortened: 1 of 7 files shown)


thousand-leaves (dismissed):

packages/webapp/src/views/common/applications.html
```diff
@@ -53,6 +53,9 @@ <h1 class="govuk-heading-xl">{{ pageHeading }}</h1>
               {% endfor %}
             </tbody>
           </table>
+        {%if showAmendmentGuidanceLink %}
```
StuAA78, line 56: For consistency, should this be `{% if` as this is how we've written it elsewhere?

```suggestion
        {% if showAmendmentGuidanceLink %}
```

StuAA78 (approved):
👍

shaundefra:
tested by Leena

PR #800, opened 2024-07-03
BNGP-5193: Upgrade GOV UK Design System to v5

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5193

This PR covers breaking and recommended changes as documented in the [changelog](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md). Amending task lists to use the new `govukTaskList` component is covered in https://github.com/DEFRA/biodiversity-net-gain-service/pull/813

Changes:
- Update dependencies
- Update paths as required by new version
- Update `Beta` phase banner to correct capitalisation as per v5 style
- Update CSP hash
- Remove deprecated classes, attributes, tags
- Remove IE8 support features (no longer supported in v5)
- Update new crown logo as required

PR #813, opened 2024-07-18
BNGP-5255 Change task list to v5 GOV UK Design System component

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5255

[We previously upgraded the GOV UK Design System from v4 to v5](https://github.com/DEFRA/biodiversity-net-gain-service/pull/800). The work done in the upgrade was to resolve breaking and recommended changes. v5 introduced a new [task list component](https://design-system.service.gov.uk/components/task-list/) with some design changes. We wish to use this but we felt the changes were outside the scope of the initial upgrade PR. We therefore change our existing task lists to use the new component in this separate PR.


thousand-leaves (approved):
checked locally and works as expected

shaundefra:
tested locally

PR #817, opened 2024-07-24
BNGP-5241: Add row number to metric extraction

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5241

This change adds the row number to the metric extraction data my mapping the row number to the data item using sheet_to_json.


jd-barnard:
Replaced by https://github.com/DEFRA/biodiversity-net-gain-service/pull/825

PR #821, opened 2024-07-25
BNGP-5301 Use habitat matching logic in combined case match-habitats route

jd-barnard:


thousand-leaves (dismissed):

shaundefra:
tested locally on 5260 https://github.com/DEFRA/biodiversity-net-gain-service/pull/812

thousand-leaves (approved):

shaundefra:
tested locally

PR #822, opened 2024-07-26
BNGP-5299 Combined case check and submit page

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5299


StuAA78:
On hold pending confirmation of the final page content.

PR #823, opened 2024-07-26
BNGP-5307 Update gov service header to  "Manage biodiversity gains" on all bng journeys

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5307

To ensure we're consistently following GOV UK design guidelines, we update the service header to display the same name across all pages.

The service name can be configured by changing the `SERVICE_NAME` environment variable. This defaults to `Manage biodiversity gains`.


thousand-leaves (dismissed):

shaundefra (approved):
authorised after test, updating with master branch and no other significant changes

shaundefra:
tested locally

PR #824, opened 2024-07-30
BNGP-5137: Combined case confirmation page

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5137

This change adds a confirmation page for the combined case journey. It does this by creating a new route and constants but reusing the common application-submitted view page.


jd-barnard (approved):

shaundefra:
closed as included in branch https://github.com/DEFRA/biodiversity-net-gain-service/pull/829

PR #825, opened 2024-07-30
BNGP-5241 Add row number to metric extraction

jd-barnard:


jd-barnard:
No changes as introduced by 5260

PR #828, opened 2024-08-01
BNGP-5303 validation: add error alert to match habitats before continuing

jd-barnard:


thousand-leaves (commented):
Page shows an error message for the radio buttons;

Screenshot 2024-08-02 at 09 46 17

But on other pages we use another error at the top of the page with the text 'There is a problem' and a link. Should this page be in the same style?

Screenshot 2024-08-02 at 09 45 29

thousand-leaves (approved):

thousand-leaves (approved):

shaundefra:
tested locally

PR #836, opened 2024-08-19
BNGP-5356: Update CC confirmation page

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5356

This change adds the combined case confirmation page route and links to the shared view. It also adds the “discharge planning section”.


shaundefra:
tested by leena

PR #839, opened 2024-08-22
BNGP-5358: Fix combined case validation error

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5358

Submitting a combined case application when the applicant's role is land owner causes an error. This is due to a mismatch between the structure of the data and the validator. After confirming that the data is structured correctly, we amend the validator to match the expected structure.


paulsimonandrews (dismissed):
LGTM

shaundefra:
tested locally

PR #840, opened 2024-08-23
BNGP-5362: Fix incorrect flow logic when changing from Check Your Answers

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5362

Once a user has gone through the "register land" or "combined case" journey and arrived at the Check Your Answers page, there is an issue with some of the Change sections where users are incorrectly returned to the task list instead of Check Your Answers after completing a journey.

The Change links for the following options present the user with an "Are you sure you want to change...?" page:

- Acting on behalf of a client
- Client is an individual or organisation
- Applicant is an individual or an organisation

(Note that "Type of legal agreement" also presents the user with 

If the user selects "No", we would expect them to be returned to Check Your Answers. Instead, they are shown a mini Check Your Answers page, and continuing from there takes the user back to the task list.

If the user selects "Yes", we would expect them to be presented with a series of questions, followed by a mini Check Your Answers page, and then be returned to Check Your Answers. Instead, they go through the questions, are presented with a mini Check Your Answers page, and continuing from there takes the user back to the task list.

This PR fixes these in two ways:

- To fix the "No" scenario, we check to see if the referrer is a Check Your Answers page; if so, they are returned there. Otherwise, we proceed to the referrer or next url as before.
- To fix the "Yes" scenario, we can't rely on the referrer since it is cleared in some scenarios. We therefore introduce a new Redis key `CHECK_AND_SUBMIT_JOURNEY_ROUTE`. When the user visits a Check Your Answers page the key is set to the page's url. The mini Check Your Answers pages have their logic amended to use this url if one is set.

### DIFF
--- a/packages/webapp/src/journey-validation/shared/applicant-info.js
+++ b/packages/webapp/src/journey-validation/shared/applicant-info.js
@@ -236,96 +236,57 @@ const checkAppInfoRoute = (startUrl, nextUrl) => routeDefinition(
   }
 )
 
-const changeClientIndividualOrganisationRoute = (startUrl, nextUrl, nextUrl1) => routeDefinition(
+const changeRoute = (payloadKey, errorText, startUrl, nextUrlYes, nextUrlNo, additionalKeysToClear = []) => routeDefinition(
   startUrl,
   [],
   (session, request) => {
-    const { changeClientIndividualOrganisation } = request.payload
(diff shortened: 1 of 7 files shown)


thousand-leaves (approved):

shaundefra:
tested locally

PR #844, opened 2024-09-02
BNGP-5381: Allocation metric section should not show completed if registration metric is changed

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5381

If a combined case user changes the registration metric file then any matched habitats are no longer applicable. Currently, changing the file does not mark the "Match available habitats" as not being completed. This PR fixes this.

While checking exactly which sections should be marked as not completed, it was confirmed that the "Confirm development and habitat details" section should be named "Confirm development details", so we make this change here as well.


thousand-leaves (approved):

paulsimonandrews (approved):
LGTM

paulsimonandrews (approved):
LGTM

shaundefra:
tested locally

PR #846, opened 2024-09-03
BNGP-5357: Metric file should not delete until new upload

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5357

This change fixes a bug that prematurely deletes the metric file data. We fix this by removing the deletion logic from the checkDeveloperUploadMetric function and moving it to the uploadAllocationMetric post request.


Stewart-Jumbe (approved):
Looks good

shaundefra:
tested locally

PR #849, opened 2024-09-05
BNGP-5331: Fix back button on combined case Defra account causing error

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5331

During a combined case application, opting to change Defra account details from the mini Check Your Answers page and then hitting the back button incorrectly takes the user back to the task list, and attempting to go back into the "Add details about the applicant" section causes a 500 error.

On investigation it turns out that the user is being taken back to the _land_ task list; we therefore fix the issue by introducing a check in the "Add details about the applicant" section to see if this is land or combined case and return the user to the appropriate task list.


thousand-leaves (approved):

shaundefra:
tested locally

PR #851, opened 2024-09-05
BNGP-5422: Fix combined case back button redirecting to land task list

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5422

Some pages are shared between the `/land` and `/combined-case` journeys. If mandatory data is missing when the page is accessed, the user is redirected to the `/land` task list, regardless of whether they were completing a `/land` or a `/combined-case` application. If they then try to go back into that section from the `/land` task list and attempt to complete it, a 500 error will be thrown.

We fixed this for `/combined-case/check-applicant-information` in [a previous PR](https://github.com/DEFRA/biodiversity-net-gain-service/pull/849); we've since identified 3 more pages with this issue so we resolve them in this PR:

```
/combined-case/check-legal-agreement-details
/combined-case/check-metric-details
/combined-case/check-ownership-proof-file
```

### DIFF
--- a/packages/webapp/src/routes/__tests__/land/check-legal-agreement-details.spec.js
+++ b/packages/webapp/src/routes/__tests__/land/check-legal-agreement-details.spec.js
@@ -49,7 +49,7 @@ describe('Legal Agreement controller tests', () => {
       })
     })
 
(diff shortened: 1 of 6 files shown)


thousand-leaves (approved):

shaundefra:
tested locally

PR #854, opened 2024-09-09
BNGP-5091: Focus on updated field when returning to check and submit page

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5091

When a user returns to the combined case check and submit or task list pages, we want the page to focus on the answer they just changed. In order to do this, this PR implements `answerIdHandler` which hooks into the Hapi lifecycle to do the following:

- Check if the current request has `journey-start-answer-id` set and if so, adds it to a call stack in the session;
- Check if the currently-requested page is a task list or check and submit, and if the call stack contains an id, it pops the id from the stack and adds `#id-of-answer` to the request so that it focuses on the answer they just changed.

We update the task list so that for specifically enabled journeys, we add the `journey-start-answer-id` query param to each Change link of "tasklist-like" pages (ie. ones which generate their content using `task-list-generator.js`).

To enable this for other journeys in future:

- Amend `task-list-generator.js` to pass `true` for the `enableJourneyStart` parameter of `generateTaskList()` for the journey's task list.
- Amend the check and submit page to use `{{ generateSummaryList }}` macro in place of `{{ govukSummaryList }}`. This requires a straightforward change to the structure of the data, from:
```
{
  key: {
    text: "File uploaded"
  },
  value: {
    html: "<span data-testid=\"land-boundary-file-name-value\">" + landBoundaryFileName + "</span>"
  },
  actions: {
    items: [
      {
        href: checkLandBoundaryLink,
        text: "Change",
        visuallyHiddenText: " land boundary file"
      }
    ]
  }
}
```
to:
```
{
  text: 'File uploaded',
  value: landBoundaryFileName,
  valueDataTestId: 'land-boundary-file-name-value',
  valueId: 'land-boundary-file-name-value',
  visuallyHiddenText: ' land boundary file',
  href: checkLandBoundaryLink,
  show: true
}
```
(Note the addition of `valueId`, which can simply be the same value as `valueDataTestId`)
- Add the task list, check and submit, and any other pages that implement the task list component to `constants.answerIdRoutes` to enable the functionality for the pages.
- Add the journey start page (ie. the one that shows the user's current applications and has the green button to start a new application) to `constants.answerIdClearRoutes`. This stops the user being incorrectly taken to a stale answer if they "break out" of the journey by starting to change an answer and then click the service name at the top.

Once it is enabled for other journeys, code that enables it to be optional for a journey can be removed:
- When it's enabled for the land journey, `get-land-check-and-submit-details.js` can be updated to remove `landownershipFilesRows` as this is no longer needed. If desired, `landownershipFilesRowsForGenerateSummaryList` can be renamed to `landownershipFilesRows` (with the check and submit pages being updated as appropriate)
- When it's enabled for all journeys, `task-list-generator.js` can be updated to remove the now-unneeded `enableJourneyStart` parameter.


thousand-leaves (dismissed):

thousand-leaves (approved):

shaundefra:
tested locally

PR #859, opened 2024-09-13
BNGP-5437: Add Qualtrics links to Combined Case

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5437

We add the required Qualtrics links to combined case header and submitted page.


paulsimonandrews (changes requested):
My bad, can you remove the `BNG` from the end of both the links?

PR #860, opened 2024-09-13
BNGP-5438: Fix match available habitats flow

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5438

This change addresses the issue where users are not redirected to the mini CYA (Check Your Answers) page immediately after confirming the developer metric file in the combined case journey after they have successfully matched habitats. We do this by adding a check to see if habitats are already matched and then changing the routing accordingly.


StuAA78 (approved):
👍

PR #861, opened 2024-09-16
BNGP-5439: Fix add landowner organisation 500 error

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5439

Attempting to add an additional landowner organisation for a combined case application results in a 500 error. This is due to the `POST` handler of `add-landowner-organisation` sending the user to the `land/check-landowners` route regardless of the application type. We fix this by adding a check to see if this is a combined case application and sending the user to `combined-case/check-landowners` if so.


thousand-leaves (approved):

shaundefra:
tested locally

PR #863, opened 2024-09-17
BNGP-5443: Add service menu to error pages

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5443

The generic error message pages do not allow the user to sign out or go directly to the application dashboard directly from it. We fix this by removing the `hideSigninNav` toggle from the 500 and 404 pages. We also add logic to pass the username to the 500 page and handle unauthorised 500's.


StuAA78:
This PR has been superseded by https://github.com/DEFRA/biodiversity-net-gain-service/pull/868 so we can close it without merging.

PR #866, opened 2024-09-18
BNGP-5448: Fix metric task list state being cleared before new file uploaded

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5448

If a user uploads a statutory biodiversity metric file, then goes to upload a new one but returns to the task list before actually uploading it, the statuses of "Add habitat baseline, creation and enhancements" and "Match available habitats" are being reset.

We fix the first by no longer using `METRIC_FILE_CHECKED` (which is used to determine routing after a user selects "yes, the file is okay" or "no, upload a new file") as a session key to determine the status. Note that this bug was raised for combined case but it also affects the land journey, so we ensure the fix is applied here too.

We fix the second by moving the clearing of the relevant session keys to after a file is successfully uploaded.


thousand-leaves (approved):

shaundefra:
tested locally

PR #867, opened 2024-09-18
BNGP-5456: prevent save of application on combined case submission

paulsimonandrews:


thousand-leaves (approved):
tested successfully

shaundefra:
tested locally

PR #868, opened 2024-09-19
BNGP-5442: Show error if metric validation fails on submission

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5442

Currently, when a user submits an application with a metric error, a generic 500 error page is displayed. This doesn't inform the user of the actual issue. To improve the user experience, we need to display a clear error message indicating that the metric file has an error and prompting the user to fix and re-upload the file.


StuAA78 (approved):
👍

shaundefra:
tested locally

PR #878, opened 2024-09-24
BNGP-5465: Revert "BNGP-5455: Tweak combined case payload to match backend processing"

paulsimonandrews:
Reverts DEFRA/biodiversity-net-gain-service#865


shaundefra:
tested locally

PR #882, opened 2024-09-26
BNGP-5467: Fix `DAC_Headings_01`

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5467

This change fixes the issue raised in the Accessibility report: DAC_Headings_01 where we have an unwanted h1 tag in a fieldset. We fix it by changing the `isPageHeading` from `true` to `false`


StuAA78 (approved):
🙌

shaundefra:
tested locally

PR #883, opened 2024-09-26
BNGP-5476: Fix other landowners flow

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5476

This change fixes the flow of the `any other landowners` section in the `legal agreement` section. It fixes for both registration and combined case journeys by updating the `altNextUrl` path in the journey validation.


StuAA78 (approved):
👍

shaundefra:
tested locally

PR #891, opened 2024-10-04
BNGP-5459: Add timestamp to applications pages

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5459

This changes adds a timestamp to the last updated field on all of the applications listing pages. This will help the user differentiate between applications submitted on the same day. We do this by updating and renaming the `getFormattedDate` function to also show the formatted time.


StuAA78 (approved):
👍

shaundefra:
tested locally

PR #893, opened 2024-10-07
BNGP-5489: Fix user is incorrectly returned from "Add legal agreement" journey

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5489


shaundefra:
tested locally

PR #901, opened 2024-10-16
BNGP-5510: Fix `answer-id-handler` to redirect to relative paths

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5510

We encountered an issue in some environments due to the `answer-id-handler` plugin redirecting to fully-qualified URLs in https://eaflood.atlassian.net/browse/BNGP-5091. To fix this we amend it so that it redirects to relative paths.


paulsimonandrews (approved):
LGTM

shaundefra:
tested locally with no visible regressions

PR #902, opened 2024-10-18
BNGP-5506: Last updated time is an hour out

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5506

This change fixes the issue of the last updated time for applications being displayed an hour off. The issue was caused by the `getFormattedDateTime` function using UTC time. We've fixed it by updating the helper function to use the built-in JavaScript `toLocaleString` method instead of moment.


StuAA78 (approved):

shaundefra:
tested locally

PR #905, opened 2024-10-22
BNGP-5505: fix uploaded file cannot be downloaded

StuAA78:
https://eaflood.atlassian.net/browse/BNGP-5505

After a user uploads a file, the "Is this the correct file?" confirmation page provides a link that allows them to re-download the file. This page is also accessible later from the task list and from check your answers.

On some pages, this link doesn't work as a route for it hasn't been defined. This PR creates the missing routes.

It also fixes the download page for the habitat plan file, as this is broken due to it pointing to the wrong blob storage container; and fixes the download for land ownership details, as this is broken due to the blob name not being set in the session.


thousand-leaves (approved):

shaundefra:
tested locally

PR #906, opened 2024-10-29
BNGP-5512: Duplicate file upload results in missing file

thousand-leaves:
https://eaflood.atlassian.net/browse/BNGP-5512

Uploading a file with the same name multiple times causes the file to be unavailable for download from the payload and browser page. The bug affects most upload pages on the registration (and therefore the combined case) journeys. 
We fix the bug by writing a new `deleteBlobFromContainersWithCheck` function. This new function ensures that blobs are only deleted if they are different from the current blob, preventing unnecessary deletions.

### Utility Function Update:
* Introduced `deleteBlobFromContainersWithCheck` in `azure-storage.js`:
  * Added the new function to ensure blobs are only deleted if they are different from the current blob.

### Route Handlers Update:
* Updated route handler files to use `deleteBlobFromContainersWithCheck`:
  * `upload-habitat-plan.js`
  * `upload-land-boundary.js`
  * `upload-local-land-charge.js`
  * `upload-metric.js`
  * `upload-ownership-proof.js`
  * `upload-written-authorisation.js`

### DIFF
--- a/packages/webapp/src/routes/land/upload-ownership-proof.js
+++ b/packages/webapp/src/routes/land/upload-ownership-proof.js
@@ -13,7 +13,7 @@ const LAND_OWNERSHIP_ID = '#landOwnership'
 async function processSuccessfulUpload (result, request, h) {
   const tempFile = request.yar.get(constants.redisKeys.TEMP_LAND_OWNERSHIP_PROOF)
   if (tempFile && !tempFile.confirmed) {
-    await deleteBlobFromContainersWithCheck(tempFile.fileLocation)
+    await deleteBlobFromContainersWithCheck(result.config.blobConfig.blobName, tempFile.fileLocation)
   }
   const tempFileDetails = {
     id: generateUniqueId(),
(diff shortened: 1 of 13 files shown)


packages/webapp/src/routes/land/upload-ownership-proof.js
```diff
@@ -13,7 +13,7 @@ const LAND_OWNERSHIP_ID = '#landOwnership'
 async function processSuccessfulUpload (result, request, h) {
   const tempFile = request.yar.get(constants.redisKeys.TEMP_LAND_OWNERSHIP_PROOF)
   if (tempFile && !tempFile.confirmed) {
-    await deleteBlobFromContainers(tempFile.fileLocation)
+    await deleteBlobFromContainersWithCheck(tempFile.fileLocation)
```
StuAA78, line 16: Only one blob path is being passed here. My reading of the code is that it should be:
```js
await deleteBlobFromContainersWithCheck(result.config.blobConfig.blobname, tempFile.fileLocation)
```

But I could be wrong here!

StuAA78 (approved):
🙌

shaundefra:
tested locally

PR #909, opened 2025-03-03
Migration to drop postgis-topology extension

paulsimonandrews:


docker/infrastructure.yml
usman-j-shaikh, line 1: @paulsimonandrews 
I am experiencing issues with Docker volume compatibility while using PostgreSQL version 16.3

usman-j-shaikh (approved):
LGTM

